PR #193, opened 2026-05-17
docs: clarify INSTALL_DIR and RIMBA_QUIET scope

wyf027:
Documents INSTALL_DIR as an install.sh shell variable and narrows RIMBA_QUIET to optional hints only.

Fixes #175

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,12 @@
 curl -sSfL https://raw.githubusercontent.com/lugassawan/rimba/main/scripts/install.sh | bash
 ```
 
+The install script honors the shell variable `INSTALL_DIR` (default: `$HOME/.local/bin`). This is read by `scripts/install.sh`, not by the `rimba` binary:
+
+```sh
+INSTALL_DIR="$HOME/bin" curl -sSfL https://raw.githubusercontent.com/lugassawan/rimba/main/scripts/install.sh | bash
+```
+
 ### Go install
 
 ```sh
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -60,6 +60,12 @@
 curl -sSfL https://raw.githubusercontent.com/lugassawan/rimba/main/scripts/install.sh | bash
 ```
 
+The install script honors the shell variable `INSTALL_DIR` (default: `$HOME/.local/bin`). This is read by `scripts/install.sh`, not by the `rimba` binary:
+
+```sh
+INSTALL_DIR="$HOME/bin" curl -sSfL https://raw.githubusercontent.com/lugassawan/rimba/main/scripts/install.sh | bash
```
lugassawan, line 66: **Critical: Shell syntax fails to pass `INSTALL_DIR` into the piped subshell**

`VAR=value command` sets the variable only in the environment of `command` — here `curl`. The `bash` process on the right-hand side of the pipe is a separate child process that does not inherit the prefix; it never sees `INSTALL_DIR`, so the script silently installs to the default `$HOME/.local/bin`.

Correct single-liner:
```sh
INSTALL_DIR="$HOME/bin" bash -c 'curl -sSfL https://raw.githubusercontent.com/lugassawan/rimba/main/scripts/install.sh | bash'
```

Or as two explicit steps:
```sh
export INSTALL_DIR="$HOME/bin"
curl -sSfL https://raw.githubusercontent.com/lugassawan/rimba/main/scripts/install.sh | bash
```

lugassawan (commented):
## Review Summary

The PR addresses both acceptance criteria from issue #175. The `INSTALL_DIR` variable name is correct and verified against `scripts/install.sh:6`. The `RIMBA_QUIET` description is substantially more accurate, correctly naming both suppressed categories (pre-execution flag hints and post-update tips) and what is not suppressed. One issue requires a fix before merge: the shell one-liner in the README uses incorrect syntax that silently fails to pass `INSTALL_DIR` into the `curl | bash` subprocess on most POSIX shells. Everything else is sound.

Detailed feedback in inline comments.

**Review Decision: COMMENT**

---
_Reviewed by `reviewer` ([swe-workbench](https://github.com/lugassawan/swe-workbench)). Posted 1 inline comment, deduped 0._

PR #196, opened 2026-05-18
[chore] Apply errhint.WithFix to remaining bare fmt.Errorf sites

lugassawan:
## Summary

- Wrap 9 bare `fmt.Errorf` sites in `cmd/init.go` with `errhint.WithFix`; extract `localConfigHint` and `gitignoreHint` package-level consts for repeated messages
- Hoist hand-rolled `\nTo recover:` / `\nTo retry:` strings out of `internal/operations/rename.go` into `errhint.WithFix`; wrap the branch-already-exists site; augment existing unit tests and e2e cases to assert `"To fix:"` + hint fragment
- Wrap 13 bare `fmt.Errorf` sites in `cmd/update.go` (network, binary, PATH stages); extract `retryUpdateHint` const; no new tests — orchestration lacks FS/process seams (tracked in #195)

## Test Plan

- [ ] E2E tests pass (`make test-e2e`)

## Issue

Closes #172

### DIFF
--- a/cmd/init.go
+++ b/cmd/init.go
@@ -222,7 +222,7 @@ func runInitGlobal(cmd *cobra.Command, uninstall bool) error {
 	if err != nil {
 		return errhint.WithFix(
 			fmt.Errorf("resolve home directory: %w", err),
-			"set HOME to your user home directory: export HOME=/Users/<you>",
+			"set HOME to your user home directory (e.g. /home/<username> on Linux, /Users/<username> on macOS)",
 		)
 	}
 	if uninstall {
@@ -245,7 +245,7 @@ func runInstall(
 	if err != nil {
 		return errhint.WithFix(
 			fmt.Errorf("agent files: %w", err),
-			"check write permissions for the install dir, or retry: rimba init --agents",
+			"check write permissions for the install dir",
 		)
 	}
 	var mcps []agentfile.Result
@@ -254,7 +254,7 @@ func runInstall(
 		if err != nil {
 			return errhint.WithFix(
 				fmt.Errorf("mcp servers: %w", err),
-				"check write permissions for MCP client configs (.mcp.json, .cursor/mcp.json, ~/.claude/settings.json), or retry: rimba init --agents",
+				"check write permissions for MCP client configs (.mcp.json, .cursor/mcp.json, ~/.claude/settings.json)",
 			)
 		}
 	}
--- a/cmd/init.go
+++ b/cmd/init.go
@@ -29,6 +29,11 @@ const (
 	tierProjectLocal installTier = "project-local"
 )
 
+const (
+	localConfigHint = "check directory permissions for .rimba/ in the repo root"
+	gitignoreHint   = "check write permissions for .gitignore in the repo root"
+)
+
 var initCmd = &cobra.Command{
 	Use:   "init",
 	Short: "Initialize rimba in the current repository",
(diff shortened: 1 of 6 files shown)


cmd/init.go
```diff
@@ -134,18 +139,21 @@ func runInitFresh(cmd *cobra.Command, r git.Runner, repoRoot, dirPath, gitignore
 
 	if !personal {
 		if err := os.WriteFile(filepath.Join(dirPath, config.LocalFile), nil, 0600); err != nil {
```
lugassawan, line 141: **Medium** — `config.Save` (2 lines below) is the only error-returning call in `runInitFresh` not wrapped with `errhint.WithFix`. Every adjacent failure in this function uses `localConfigHint`, but a permission failure on `.rimba/settings.toml` silently omits the hint.

```go
if err := config.Save(filepath.Join(dirPath, config.TeamFile), cfg); err != nil {
    return errhint.WithFix(fmt.Errorf("failed to save team config: %w", err), localConfigHint)
}
```

internal/operations/rename.go
```diff
@@ -39,11 +43,18 @@ func RenameWorktree(r git.Runner, wt resolver.WorktreeInfo, newTask, wtDir strin
```
lugassawan, line 46: **Medium** — the forward `git.MoveWorktree` call (2 lines above this hunk) returns a bare error. Both rollback branches in this function already receive `errhint.WithFix` hints; the forward-move path is equally user-facing (e.g. locked worktree without `--force`).

`TestRenameWorktreeMoveFails` also only asserts `err != nil` — no hint assertion.

Suggested fix at the `MoveWorktree` return:
```go
if err := git.MoveWorktree(r, wt.Path, newPath, force); err != nil {
    return RenameResult{}, errhint.WithFix(
        fmt.Errorf("failed to move worktree: %w", err),
        "unlock the worktree if locked: git worktree unlock "+wt.Path+", then retry: rimba rename",
    )
}
```

cmd/init.go
lugassawan: Addressed in 05c237db1f478d24da9326ac874d94c990365e66: wrapped config.Save with errhint.WithFix + localConfigHint.

internal/operations/rename.go
lugassawan: Addressed in 05c237db1f478d24da9326ac874d94c990365e66: wrapped forward git.MoveWorktree with errhint.WithFix; updated TestRenameWorktreeMoveFails to assert hint fragment.

cmd/init.go
```diff
@@ -229,13 +243,19 @@ func runInstall(
 ) error {
 	files, err := installFn(dir)
 	if err != nil {
-		return fmt.Errorf("agent files: %w", err)
+		return errhint.WithFix(
+			fmt.Errorf("agent files: %w", err),
+			"check write permissions for the install dir, or retry: rimba init --agents",
```
lugassawan: **Low** — when `rimba init --agents --uninstall` fails, the hint says "retry: rimba init --agents" which re-installs rather than retries the removal. The same `runInstall` function handles both install and uninstall, so the hint is always phrased as an install retry.

Consider either passing a verb-specific hint through `runInstall` or rephrasing to just: `"check write permissions for the install dir"` (omit the retry command to avoid misdirection).

cmd/init.go
```diff
@@ -209,7 +220,10 @@ func runInitAgents(cmd *cobra.Command, repoRoot string, local, uninstall bool) e
 func runInitGlobal(cmd *cobra.Command, uninstall bool) error {
 	home, err := os.UserHomeDir()
 	if err != nil {
-		return fmt.Errorf("resolve home directory: %w", err)
+		return errhint.WithFix(
+			fmt.Errorf("resolve home directory: %w", err),
+			"set HOME to your user home directory: export HOME=/Users/<you>",
```
lugassawan: **Low** — the HOME hint `export HOME=/Users/<you>` hard-codes a macOS path. Linux users have `/home/<username>` and Windows uses a different env var entirely.

Consider a platform-agnostic phrasing such as:
```suggestion
		"set HOME to your user home directory (e.g. /home/<username> on Linux, /Users/<username> on macOS)",
```

cmd/init.go
lugassawan: Addressed in a3d7a8a: dropped the "retry: rimba init --agents" clause from both `runInstall` error hints — the retry command is caller-dependent (install vs uninstall vs local vs global), so hinting a specific command is misleading. The permission check is now the standalone actionable hint.

cmd/init.go
lugassawan: Addressed in a3d7a8a: replaced the macOS-specific `export HOME=/Users/<you>` with a platform-agnostic phrasing: `"set HOME to your user home directory (e.g. /home/<username> on Linux, /Users/<username> on macOS)"`.

cmd/update.go
```diff
@@ -56,41 +62,56 @@ refresh). Set RIMBA_QUIET=1 to suppress the tip.`,
...
 			}
 
 			// Install to user-writable directory instead
 			s.Stop()
 			userDir, dirErr := updater.UserInstallDir()
 			if dirErr != nil {
-				return fmt.Errorf("getting user install dir: %w", dirErr)
+				return errhint.WithFix(
+					fmt.Errorf("getting user install dir: %w", dirErr),
+					"set HOME or XDG_DATA_HOME and retry: rimba update",
```
lugassawan: **Low** — the hint mentions `XDG_DATA_HOME` but `internal/updater.UserInstallDir()` only calls `os.UserHomeDir()` and constructs `~/.local/bin` from that — it never reads `XDG_DATA_HOME`. A user who follows the hint, sets `XDG_DATA_HOME`, and retries will get the same error because the variable is silently ignored.

```suggestion
					"set HOME to your user home directory and retry: rimba update",
```

lugassawan: Addressed in daa3784: dropped `XDG_DATA_HOME` from the hint — `UserInstallDir()` only calls `os.UserHomeDir()`, which reads `$HOME` on Linux/macOS. Hint is now `"set HOME to your user home directory and retry: rimba update"`.

PR #200, opened 2026-05-18
[ci] Workflow improvements: cache, dedupe, coverage timeout

lugassawan:
## Summary
- Enable Go module caching (`cache: true`) on all three `actions/setup-go@v6` steps, eliminating redundant module downloads on every job run
- Remove the duplicate `test` job — `coverage` already runs the full suite via `make test-coverage`; parallel wall-clock time is unchanged
- Add `timeout-minutes: 10` to the `coverage` job to guard against hung e2e worktree operations blocking a runner indefinitely

## Test Plan
- [x] E2E tests pass (`make test-coverage` — 97.7% coverage, threshold 97%)
- [ ] CI run shows 3 jobs (`lint`, `coverage`, `build`) instead of 4
- [ ] `Set up Go` step logs show `Cache restored from key: ...` on second push (cache miss expected on first run)
- [ ] `coverage` job summary shows `timeout-minutes: 10`

## Issue

Closes #178

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,10 +15,9 @@ jobs:
         with:
           go-version: stable
           cache: true
-      - uses: golangci/golangci-lint-action@v6
+      - uses: golangci/golangci-lint-action@v7
         with:
           version: v2.11.4
-          args: --config=.golangci.yml ./...
 
   coverage:
     runs-on: ubuntu-latest
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -7,15 +7,6 @@ on:
     branches: [main]
 
 jobs:
-  test:
-    runs-on: ubuntu-latest
-    steps:
-      - uses: actions/checkout@v6
-      - uses: actions/setup-go@v6
-        with:
-          go-version: stable
-      - run: make test
-
   lint:
     runs-on: ubuntu-latest
     steps:
@@ -23,16 +14,19 @@ jobs:
       - uses: actions/setup-go@v6
         with:
           go-version: stable
+          cache: true
       - run: go install github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.11.4
       - run: make lint
 
   coverage:
     runs-on: ubuntu-latest
+    timeout-minutes: 10
     steps:
       - uses: actions/checkout@v6
       - uses: actions/setup-go@v6
         with:
           go-version: stable
+          cache: true
       - name: Run tests with coverage
         run: make test-coverage
       - name: Check coverage threshold
@@ -52,4 +46,5 @@ jobs:
       - uses: actions/setup-go@v6
         with:
           go-version: stable
+          cache: true
       - run: make build
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -15,9 +15,12 @@ jobs:
         with:
           go-version: stable
           cache: true
-      - uses: golangci/golangci-lint-action@v7
+      - uses: actions/cache@v4
         with:
-          version: v2.11.4
+          path: ~/go/bin/golangci-lint
+          key: golangci-lint-v2.11.4
+      - run: go install github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.11.4
+      - run: make lint
 
   coverage:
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -7,32 +7,26 @@ on:
...
-
   lint:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v6
       - uses: actions/setup-go@v6
         with:
           go-version: stable
+          cache: true
       - run: go install github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.11.4
```
lugassawan: **Medium — golangci-lint binary is not cached**

`go install` re-downloads and rebuilds the golangci-lint binary on every run. The `cache: true` on `actions/setup-go` only caches project-module downloads (keyed on `go.sum`); installed binaries in `~/go/bin` are not covered.

On a warm module-cache run, the `go install` step will still re-download golangci-lint (a ~60 MB binary) from the network each time, partially negating the wall-clock savings.

**Suggested fix:** switch to the official `golangci/golangci-lint-action`, which ships its own binary cache:
```yaml
- uses: golangci/golangci-lint-action@v6
  with:
    version: v2.11.4
```
Or add an explicit `actions/cache` step keyed on the version string.

.github/workflows/ci.yml
```diff
@@ -7,32 +7,26 @@ on:
...
     runs-on: ubuntu-latest
+    timeout-minutes: 10
     steps:
       - uses: actions/checkout@v6
       - uses: actions/setup-go@v6
         with:
           go-version: stable
+          cache: true
       - name: Run tests with coverage
         run: make test-coverage
```
lugassawan, line 35: **Low — `timeout-minutes: 10` may be tight on a cold-cache runner**

`make test-coverage` runs `go test ./...` which includes the e2e suite in `tests/e2e/`. On a first run (cache miss), the job must compile the module, build the instrumented binary (`go build -cover`), spin up real git worktrees, and run the full e2e suite — all within 10 minutes.

The Makefile's `test-e2e` target already sets a per-package `-timeout 120s`, but that only covers individual e2e test execution time, not overall compilation + startup overhead.

This is unlikely to flap given that local verification passed at 97.7%, but worth monitoring on the first few CI runs. **Suggested fix:** bump to `timeout-minutes: 15` initially, then tighten once you have a few green-run baselines establishing p95 wall-clock.

.github/workflows/ci.yml
lugassawan: Addressed in 7186588: replaced `go install` + `make lint` with `golangci/golangci-lint-action@v6` (version: v2.11.4), which caches the compiled binary across runs.

.github/workflows/ci.yml
lugassawan: Addressed in 7186588: bumped `timeout-minutes` from 10 to 15 to give headroom on cold-cache runners. Will tighten based on observed p95 once a baseline is established.

.github/workflows/ci.yml
```diff
@@ -7,32 +7,30 @@ on:
...
       - uses: actions/checkout@v6
       - uses: actions/setup-go@v6
         with:
           go-version: stable
+          cache: true
+      - uses: actions/cache@v4
+        with:
+          path: ~/go/bin/golangci-lint
+          key: golangci-lint-v2.11.4
       - run: go install github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.11.4
```
lugassawan: **Important — golangci-lint pinned at stale version**

The version string appears in two places that must stay in sync:
- Line 21: cache key `golangci-lint-v2.11.4`
- Line 22: `go install ...@v2.11.4`

Latest release as of 2026-05-18 is `v2.12.2` (shipped 2026-05-06). The current pin is valid and CI passes, but it silently misses linter fixes and new checks from v2.12.x.

**Suggested fix:** bump both atomically to `v2.12.2`:
```yaml
          key: golangci-lint-v2.12.2
      - run: go install github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.12.2
```

lugassawan: Addressed in d204e7f: bumped both the cache key and go install pin from v2.11.4 to v2.12.2 atomically.

PR #201, opened 2026-05-19
[docs] cobra/cmd UX sweep: Example fields, restore help, version Go runtime, flag casing

lugassawan:
## Summary

- Add `go:` line (`runtime.Version()`) to `rimba version` output — matches de-facto Go CLI convention
- Expand `rimba restore --help` to document the `rimba archive` prerequisite and the `"no archived branch found for task <name>"` failure case
- Lowercase all flag `Usage` strings across `cmd/` to match Go stdlib `flag` convention
- Add `Example:` field to every runnable cobra command; extract examples from `Long:` where they were embedded (`add`, `list`, `open`, `status`, `completion`)
- Add `TestFlagUsageIsLowercase` and `TestExamplePresent` metatests to guard against future drift

## Test Plan

- [x] `bin/rimba version` includes a `go:` line
- [x] `bin/rimba help restore` mentions `rimba archive` and the missing-branch case
- [x] `bin/rimba help <cmd>` shows an Examples section for every command

## Issue

Closes #173

### DIFF
--- a/cmd/conventions_test.go
+++ b/cmd/conventions_test.go
@@ -1,0 +1,45 @@
+package cmd
+
+import (
+	"testing"
+	"unicode"
+
+	"github.com/spf13/cobra"
+	"github.com/spf13/pflag"
+)
+
+// allCommands returns cmd and all its descendants recursively.
+func allCommands(c *cobra.Command) []*cobra.Command {
+	result := make([]*cobra.Command, 0, 1+len(c.Commands()))
+	result = append(result, c)
+	for _, sub := range c.Commands() {
+		result = append(result, allCommands(sub)...)
+	}
+	return result
+}
+
+func TestFlagUsageIsLowercase(t *testing.T) {
+	for _, c := range allCommands(rootCmd) {
+		c.LocalFlags().VisitAll(func(f *pflag.Flag) {
+			if f.Hidden || f.Usage == "" {
+				return
+			}
+			first := rune(f.Usage[0])
+			if unicode.IsUpper(first) {
+				t.Errorf("command %q flag --%s: Usage starts with uppercase %q (must be lowercase): %q",
+					c.CommandPath(), f.Name, string(first), f.Usage)
+			}
+		})
+	}
+}
+
+func TestExamplePresent(t *testing.T) {
+	for _, c := range allCommands(rootCmd) {
+		if !c.Runnable() {
+			continue
+		}
+		if c.Example == "" {
+			t.Errorf("runnable command %q has no Example field", c.CommandPath())
+		}
+	}
+}
--- a/cmd/conventions_test.go
+++ b/cmd/conventions_test.go
@@ -20,6 +20,9 @@ func allCommands(c *cobra.Command) []*cobra.Command {
 
 func TestFlagUsageIsLowercase(t *testing.T) {
 	for _, c := range allCommands(rootCmd) {
+		// LocalFlags covers this command's own Flags() and PersistentFlags(),
+		// but not flags inherited from ancestor commands — those are validated
+		// when the ancestor command is processed in the outer loop.
 		c.LocalFlags().VisitAll(func(f *pflag.Flag) {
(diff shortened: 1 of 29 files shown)


cmd/version_test.go
```diff
@@ -19,13 +19,13 @@ func TestVersionCmd(t *testing.T) {
...
-	for _, want := range []string{"rimba", version, "commit:", "built:", "os:", "arch:"} {
+	for _, want := range []string{"rimba", version, "commit:", "built:", "os:", "arch:", "go:"} {
 		if !strings.Contains(out, want) {
 			t.Errorf("version output %q does not contain %q", out, want)
 		}
 	}
 	lines := strings.Split(strings.TrimSpace(out), "\n")
-	if len(lines) != 5 {
-		t.Errorf("expected 5 lines, got %d: %q", len(lines), out)
+	if len(lines) < 6 {
```
lugassawan: **Medium:** Line-count assertion weakened without justification. The original `!= 5` was an exact regression guard; `< 6` only enforces a floor — a future change that duplicates a line or emits extra output would pass unnoticed. Since `rimba version` now emits exactly 6 lines (fixed layout), this should be an equality check:

```suggestion
	if len(lines) != 6 {
		t.Errorf("expected 6 lines, got %d: %q", len(lines), out)
	}
```

cmd/conventions_test.go
```diff
@@ -0,0 +1,45 @@
...
+	result = append(result, c)
+	for _, sub := range c.Commands() {
+		result = append(result, allCommands(sub)...)
+	}
+	return result
+}
+
+func TestFlagUsageIsLowercase(t *testing.T) {
+	for _, c := range allCommands(rootCmd) {
+		c.LocalFlags().VisitAll(func(f *pflag.Flag) {
```
lugassawan, line 26: **Low:** `LocalFlags()` excludes flags inherited from ancestor commands — the three persistent flags on `rootCmd` (`--json`, `--no-color`, `--debug`) are not re-visited on child commands; they are covered only when `rootCmd` itself is iterated. This is correct behaviour, but non-obvious. A brief comment would save the next reader a trip to the cobra source:

```suggestion
		// LocalFlags covers this command's own Flags() and PersistentFlags(),
		// but not flags inherited from ancestor commands — those are validated
		// when the ancestor command is processed in the outer loop.
		c.LocalFlags().VisitAll(func(f *pflag.Flag) {
```

cmd/restore.go
```diff
@@ -16,8 +16,16 @@ import (
 var restoreCmd = &cobra.Command{
 	Use:   "restore <task>",
 	Short: "Restore an archived worktree from its preserved branch",
-	Long:  "Recreates a worktree from a branch that was previously archived with `rimba archive`.",
-	Args:  cobra.ExactArgs(1),
+	Long: `Recreates a worktree from a branch that was previously archived with
+rimba archive. The local branch must still exist — restore does not
```
lugassawan: **Low:** "The local branch must still exist" is slightly ambiguous — it could be read as referring to the branch checked out in the current worktree. The archived branch is one that `rimba archive` left in place but detached from a worktree. "The archived branch must still exist locally" more precisely matches what `FindArchivedBranch` checks (presence in `git branch` output, not worktree attachment).

cmd/version_test.go
lugassawan: Addressed in 93607ae: tightened `len(lines) < 6` → `len(lines) != 6` to restore the two-sided regression guard.

cmd/conventions_test.go
lugassawan: Addressed in 93607ae: added a 3-line comment above the `LocalFlags().VisitAll` call documenting the ancestor-flag coverage model.

cmd/restore.go
lugassawan: Addressed in 93607ae: updated Long to "The archived branch must still exist locally" — more precisely reflects what `FindArchivedBranch` checks.

PR #202, opened 2026-05-20
[feat] hooks installer: replace make hooks with scripts/setup.sh (preflight, --force, symlinks)

lugassawan:
## Summary

- Replace `make hooks` one-liner (`git config core.hooksPath .githooks`) with `scripts/setup.sh`, a POSIX sh installer that mirrors the swe-workbench reference design.
- Preflight phase detects conflicts (custom `core.hooksPath`, existing hook files, wrong symlinks) and exits 1 without `--force`; with `--force` it warns and overwrites.
- Apply phase installs symlinks: `.git/hooks/<name> -> .githooks/<name>` — idempotent, linked-worktree-aware (GIT_DIR absolute-path normalization).
- `make hooks` becomes a thin alias; `./scripts/setup.sh` can also be invoked directly.
- CONTRIBUTING.md updated: re-runnable note, `--force` mention, previously missing `post-merge` hook description.

**Behavior-change callout for reviewers:** contributors with a custom `core.hooksPath` will now see an exit-1 conflict instead of a silent overwrite. The error message tells them to re-run with `--force`. This is intentional — not a regression.

## Test Plan

- [x] E2E tests pass (`make test-e2e`) — includes 11 new tests in `tests/e2e/setup_test.go`
- [x] Shell syntax: `sh -n scripts/setup.sh && bash -n scripts/setup.sh`

### Type-tailored checks (feat)
- [ ] Fresh clone → `./scripts/setup.sh` exits 0 and symlinks appear under `.git/hooks/`
- [ ] Re-run → exits 0, no stderr, symlinks unchanged
- [ ] `git config core.hooksPath custom-hooks` → `./scripts/setup.sh` exits 1 with conflict listed
- [ ] Same with `--force` → exits 0, `core.hooksPath` unset, symlinks correct
- [ ] `./scripts/setup.sh --help` → exits 0, prints `Usage:` and `--force`
- [ ] `./scripts/setup.sh --bogus` → exits 2, stderr contains `unknown argument`

## Issue

N/A

### DIFF
--- a/scripts/setup.sh
+++ b/scripts/setup.sh
@@ -1,0 +1,73 @@
+#!/bin/sh
+set -eu
+
+FORCE=0
+for arg in "$@"; do
+  case "$arg" in
+    --force|-f) FORCE=1 ;;
+    --help|-h)
+      printf 'Usage: %s [--force]\n' "$0"
+      printf '\n'
+      printf 'Installs symlinks in .git/hooks/ -> .githooks/.\n'
+      printf 'Pre-existing hooks and a non-default repo-local core.hooksPath\n'
+      printf 'are preserved unless --force is passed.\n'
+      exit 0 ;;
+    *) printf 'Error: unknown argument: %s\n' "$arg" >&2; exit 2 ;;
+  esac
+done
+
+ROOT="$(git rev-parse --show-toplevel)"
+GIT_DIR="$(git rev-parse --git-dir)"
+case "$GIT_DIR" in /*) ;; *) GIT_DIR="$ROOT/$GIT_DIR" ;; esac
+
+# --- Preflight: collect conflicts ------------------------------------------
+conflicts=""
+
+existing_hp="$(git -C "$ROOT" config --local --get core.hooksPath 2>/dev/null || true)"
+existing_hp="${existing_hp%/}"   # normalize trailing slash
+if [ -n "$existing_hp" ] \
+   && [ "$existing_hp" != ".githooks" ] \
+   && [ "$existing_hp" != "$ROOT/.githooks" ]; then
+  conflicts="${conflicts}  core.hooksPath = $existing_hp (expected unset or .githooks)
+"
+fi
+
+mkdir -p "$GIT_DIR/hooks"
+for src in "$ROOT"/.githooks/*; do
+  [ -e "$src" ] || [ -L "$src" ] || break   # glob unexpanded — .githooks is empty or missing
+  name="$(basename "$src")"
+  target="$GIT_DIR/hooks/$name"
+  expected="$ROOT/.githooks/$name"
+  if [ -L "$target" ]; then
+    [ "$(readlink "$target")" = "$expected" ] && continue   # already correct
+    conflicts="${conflicts}  $target -> $(readlink "$target") (expected $expected)
+"
+  elif [ -e "$target" ]; then
+    conflicts="${conflicts}  $target (regular file, not a symlink)
+"
+  fi
+done
+
+if [ -n "$conflicts" ]; then
+  if [ "$FORCE" -eq 0 ]; then
+    printf 'Error: setup.sh would overwrite existing hook state:\n' >&2
+    printf '%s' "$conflicts" >&2
+    printf 'Re-run with --force to overwrite.\n' >&2
+    exit 1
+  fi
+  printf 'warning: --force: overwriting:\n' >&2
+  printf '%s' "$conflicts" >&2
+fi
+
+# --- Apply -----------------------------------------------------------------
+case "$existing_hp" in
+  ""|.githooks|"$ROOT/.githooks") ;;   # unset or already correct — leave it alone
+  *) git -C "$ROOT" config --local --unset core.hooksPath ;;
+esac
+for src in "$ROOT"/.githooks/*; do
+  [ -e "$src" ] || [ -L "$src" ] || break   # glob unexpanded — .githooks is empty or missing
+  name="$(basename "$src")"
+  ln -sfn "$ROOT/.githooks/$name" "$GIT_DIR/hooks/$name"
+done
+
+echo "Git hooks linked: $GIT_DIR/hooks/<name> -> $ROOT/.githooks/<name>"
--- a/scripts/setup.sh
+++ b/scripts/setup.sh
@@ -72,4 +72,4 @@ for src in "$ROOT"/.githooks/*; do
   ln -sfn "$ROOT/.githooks/$name" "$GIT_DIR/hooks/$name"
 done
 
-printf 'Git hooks linked: %s/hooks/<hook> -> %s/.githooks/<hook>\n' "$GIT_DIR" "$ROOT"
+printf 'Git hooks linked: %s/hooks/ -> %s/.githooks/\n' "$GIT_DIR" "$ROOT"
--- a/scripts/setup.sh
+++ b/scripts/setup.sh
@@ -17,8 +17,7 @@ for arg in "$@"; do
 done
 
 ROOT="$(git rev-parse --show-toplevel)"
-GIT_DIR="$(git rev-parse --git-dir)"
-case "$GIT_DIR" in /*) ;; *) GIT_DIR="$ROOT/$GIT_DIR" ;; esac
+GIT_DIR="$(git -C "$ROOT" rev-parse --absolute-git-dir)"
 
 # --- Preflight: collect conflicts ------------------------------------------
 conflicts=""
@@ -35,6 +34,7 @@ fi
 mkdir -p "$GIT_DIR/hooks"
 for src in "$ROOT"/.githooks/*; do
   [ -e "$src" ] || [ -L "$src" ] || break   # glob unexpanded — .githooks is empty or missing
+  [ -L "$src" ] && [ ! -e "$src" ] && { printf 'warning: broken symlink in .githooks, skipping: %s\n' "$src" >&2; continue; }
   name="$(basename "$src")"
   target="$GIT_DIR/hooks/$name"
   expected="$ROOT/.githooks/$name"
@@ -62,12 +62,14 @@ fi
 # --- Apply -----------------------------------------------------------------
 case "$existing_hp" in
   ""|.githooks|"$ROOT/.githooks") ;;   # unset or already correct — leave it alone
-  *) git -C "$ROOT" config --local --unset core.hooksPath ;;
+  *) git -C "$ROOT" config --local --unset core.hooksPath
+     printf 'Unset core.hooksPath (was: %s)\n' "$existing_hp" ;;
 esac
 for src in "$ROOT"/.githooks/*; do
   [ -e "$src" ] || [ -L "$src" ] || break   # glob unexpanded — .githooks is empty or missing
+  [ -L "$src" ] && [ ! -e "$src" ] && { printf 'warning: broken symlink in .githooks, skipping: %s\n' "$src" >&2; continue; }
   name="$(basename "$src")"
   ln -sfn "$ROOT/.githooks/$name" "$GIT_DIR/hooks/$name"
 done
 
-echo "Git hooks linked: $GIT_DIR/hooks/<name> -> $ROOT/.githooks/<name>"
+printf 'Git hooks linked: %s/hooks/<hook> -> %s/.githooks/<hook>\n' "$GIT_DIR" "$ROOT"
--- a/tests/e2e/setup_test.go
+++ b/tests/e2e/setup_test.go
@@ -1,0 +1,364 @@
+package e2e_test
+
+import (
+	"bytes"
+	"errors"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+	"testing"
+
+	"github.com/lugassawan/rimba/testutil"
+)
+
+// conflictRow describes one conflict scenario for TestSetupConflicts.
+type conflictRow struct {
+	name  string
+	setup func(t *testing.T, repo string)
+}
+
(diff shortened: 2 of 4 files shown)


tests/e2e/setup_test.go
```diff
@@ -0,0 +1,364 @@
...
+	return p
+}
+
+// runSetup executes scripts/setup.sh inside repo with optional args, mirrors
+// the result/rimba helper pattern in e2e_test.go.
+func runSetup(t *testing.T, repo string, args ...string) result {
+	t.Helper()
+	scriptPath := resolveScriptPath(t)
+	cmdArgs := append([]string{scriptPath}, args...)
+	cmd := exec.Command("bash", cmdArgs...)
```
lugassawan: **Critical** — `runSetup` invokes the script with `bash` instead of `sh`.

The script declares `#!/bin/sh` and the PR describes it as a POSIX sh installer, but every test exercises it under bash. On Debian/Ubuntu CI, `/bin/sh` is dash; bash-isms (e.g. `[[ ]]`, `local`, arrays) would pass all these tests but fail on those systems.

**Suggested fix:** Change to `exec.Command("sh", scriptPath, args...)` to match the shebang and the portability claim.

tests/e2e/setup_test.go
```diff
@@ -0,0 +1,364 @@
...
+	mainRepo := setupRepo(t)
+
+	// Add a linked worktree
+	wtPath := filepath.Join(t.TempDir(), "linked-wt")
+	testutil.GitCmd(t, mainRepo, "worktree", "add", wtPath, "-b", "feature/wt-setup-test")
+
+	// Populate .githooks/ in the WORKTREE's working directory; when the script
+	// runs inside the worktree, ROOT = wtPath (git rev-parse --show-toplevel
+	// returns the linked worktree's root, not the main repo).
+	populateGithooks(t, wtPath)
```
lugassawan, line 317: **High** — `populateGithooks(t, wtPath)` places `.githooks/` inside the linked worktree's checkout directory. In real usage `.githooks/` lives in the **main** repo root and is tracked by git; linked worktrees don't get their own copy.

`git rev-parse --show-toplevel` inside a linked worktree returns the worktree's own directory, so the script constructs symlinks pointing at `wtPath/.githooks/name` — a path that wouldn't exist for a real contributor.

**Suggested fix:** Call `populateGithooks(t, mainRepo)` and update the expected symlink target to `mainRepo + "/.githooks/" + name`, or add a comment explicitly documenting the 'each-worktree-has-its-own-.githooks' deployment model.

scripts/setup.sh
```diff
@@ -0,0 +1,73 @@
...
+    exit 1
+  fi
+  printf 'warning: --force: overwriting:\n' >&2
+  printf '%s' "$conflicts" >&2
+fi
+
+# --- Apply -----------------------------------------------------------------
+case "$existing_hp" in
+  ""|.githooks|"$ROOT/.githooks") ;;   # unset or already correct — leave it alone
+  *) git -C "$ROOT" config --local --unset core.hooksPath ;;
```
lugassawan: **High** — when `--force` causes `core.hooksPath` to be unset, the script silently mutates git config with no stdout confirmation. The success message on line 73 says nothing about this change.

```sh
*) git -C "$ROOT" config --local --unset core.hooksPath ;;
```

A contributor running with `--force` may not realize their `core.hooksPath` was removed, especially if they set it intentionally for another purpose.

**Suggested fix:** Print a note before line 73, e.g. `printf 'Unset core.hooksPath (was: %s)\\n' "$existing_hp"`.

tests/e2e/setup_test.go
```diff
@@ -0,0 +1,364 @@
...
+		}
+		// Target is relative to the worktree root, since ROOT = wtPath.
+		expected := filepath.Join(wtPathReal, ".githooks", name)
+		if target != expected {
+			t.Errorf("worktree %s -> %q, want %q", linkPath, target, expected)
+		}
+	}
+}
+
+func TestSetupHookSourcesAreExecutable(t *testing.T) {
```
lugassawan, line 348: **Medium** — every other test in this file opens with `if testing.Short() { t.Skip(skipE2E) }`, but `TestSetupHookSourcesAreExecutable` skips that guard. The function performs real filesystem I/O against the project checkout, so `go test -short ./tests/e2e/` will always execute it, inconsistent with the rest of the e2e suite.

**Suggested fix:** Add the `testing.Short()` skip guard, or move this test to a unit test file outside the e2e package (since it only reads project source files, not a subprocess).

scripts/setup.sh
```diff
@@ -0,0 +1,73 @@
...
+existing_hp="${existing_hp%/}"   # normalize trailing slash
+if [ -n "$existing_hp" ] \
+   && [ "$existing_hp" != ".githooks" ] \
+   && [ "$existing_hp" != "$ROOT/.githooks" ]; then
+  conflicts="${conflicts}  core.hooksPath = $existing_hp (expected unset or .githooks)
+"
+fi
+
+mkdir -p "$GIT_DIR/hooks"
+for src in "$ROOT"/.githooks/*; do
```
lugassawan, line 35: **Medium** — the glob-empty guard `[ -e "$src" ] || [ -L "$src" ] || break` correctly handles unexpanded globs, but `-L $src` without `-e $src` is true for a **broken** symlink in `.githooks/`. In that case the script would still iterate and create a chain of two broken symlinks in `.git/hooks/`.

**Suggested fix:** After the existence check, add a guard to skip broken symlink sources:
```sh
[ -L "$src" ] && [ ! -e "$src" ] && { printf 'warning: broken symlink in .githooks: %s\\n' "$src" >&2; continue; }
```

scripts/setup.sh
```diff
@@ -0,0 +1,73 @@
...
+      printf 'Pre-existing hooks and a non-default repo-local core.hooksPath\n'
+      printf 'are preserved unless --force is passed.\n'
+      exit 0 ;;
+    *) printf 'Error: unknown argument: %s\n' "$arg" >&2; exit 2 ;;
+  esac
+done
+
+ROOT="$(git rev-parse --show-toplevel)"
+GIT_DIR="$(git rev-parse --git-dir)"
+case "$GIT_DIR" in /*) ;; *) GIT_DIR="$ROOT/$GIT_DIR" ;; esac
```
lugassawan: **Medium** — the `case "$GIT_DIR" in` normalization is functionally correct for normal use, but if the caller's environment has a `GIT_DIR` env var set, `git rev-parse --git-dir` echoes that env-override value regardless of the working directory. This can bite in CI environments that set `GIT_DIR` for unrelated reasons.

**Suggested fix:** Use `git -C "$ROOT" rev-parse --absolute-git-dir` (available in git 2.13+) instead:
```sh
GIT_DIR="$(git -C "$ROOT" rev-parse --absolute-git-dir)"
```
This returns an absolute path unconditionally, making the `case` normalization on line 21 unnecessary.

scripts/setup.sh
```diff
@@ -0,0 +1,73 @@
...
+  ""|.githooks|"$ROOT/.githooks") ;;   # unset or already correct — leave it alone
+  *) git -C "$ROOT" config --local --unset core.hooksPath ;;
+esac
+for src in "$ROOT"/.githooks/*; do
+  [ -e "$src" ] || [ -L "$src" ] || break   # glob unexpanded — .githooks is empty or missing
+  name="$(basename "$src")"
+  ln -sfn "$ROOT/.githooks/$name" "$GIT_DIR/hooks/$name"
+done
+
+echo "Git hooks linked: $GIT_DIR/hooks/<name> -> $ROOT/.githooks/<name>"
```
lugassawan: **Low** — all other output in the script uses `printf`, but line 73 uses `echo`. POSIX `echo` behavior with escape sequences is implementation-defined. While the string here contains no escape sequences (so it's benign today), it's inconsistent with the rest of the script.

Also, using literal `<name>` in the message rather than the actual path is less informative.

**Suggested fix:**
```sh
printf 'Git hooks linked: %s/hooks/<hook> -> %s/.githooks/<hook>\\n' "$GIT_DIR" "$ROOT"
```

tests/e2e/setup_test.go
lugassawan: Addressed in 332020820290: changed `exec.Command("bash", ...)` to `exec.Command("sh", ...)` in `runSetup` to match the `#!/bin/sh` shebang.

tests/e2e/setup_test.go
lugassawan: Good catch on the documentation gap. This test intentionally places `.githooks/` in the worktree root because `rimba add` copies tracked files (including `.githooks/`) into each linked worktree — `git rev-parse --show-toplevel` inside a linked worktree returns the worktree's own directory, so `ROOT` = `wtPath`. I've expanded the comment in the test to document this deployment model explicitly. A plain `git worktree add` checkout without that copy step would yield a no-op install (the glob expands to nothing), not a failure.

scripts/setup.sh
lugassawan: Addressed in 332020820290: added `printf 'Unset core.hooksPath (was: %s)\\n' "$existing_hp"` after the `--unset` call so the mutation is visible in stdout output.

tests/e2e/setup_test.go
lugassawan: Addressed in 332020820290: added `if testing.Short() { t.Skip(skipE2E) }` guard to `TestSetupHookSourcesAreExecutable` for consistency with the rest of the e2e suite.

scripts/setup.sh
lugassawan: Addressed in 332020820290: added broken-symlink guard in both the preflight and apply loops — `[ -L "$src" ] && [ ! -e "$src" ] && { printf 'warning: broken symlink in .githooks, skipping: %s\\n' "$src" >&2; continue; }`.

scripts/setup.sh
lugassawan: Addressed in 332020820290: replaced `git rev-parse --git-dir` + manual case normalization with `git -C "$ROOT" rev-parse --absolute-git-dir` (git 2.13+). This returns an absolute path unconditionally and is unaffected by a caller-set `GIT_DIR` env var, eliminating both the normalization case statement and the env-override edge case.

scripts/setup.sh
lugassawan: Addressed in 332020820290: changed `echo` to `printf` on the success line, and replaced the literal `<hook>` placeholder with the actual `$GIT_DIR` and `$ROOT` paths so the output shows real directory paths.

scripts/setup.sh
```diff
@@ -0,0 +1,75 @@
...
+     printf 'Unset core.hooksPath (was: %s)\n' "$existing_hp" ;;
+esac
+for src in "$ROOT"/.githooks/*; do
+  [ -e "$src" ] || [ -L "$src" ] || break   # glob unexpanded — .githooks is empty or missing
+  [ -L "$src" ] && [ ! -e "$src" ] && { printf 'warning: broken symlink in .githooks, skipping: %s\n' "$src" >&2; continue; }
+  name="$(basename "$src")"
+  ln -sfn "$ROOT/.githooks/$name" "$GIT_DIR/hooks/$name"
+done
+
+printf 'Git hooks linked: %s/hooks/<hook> -> %s/.githooks/<hook>\n' "$GIT_DIR" "$ROOT"
```
lugassawan: **Low** — the fix replaced `echo` with `printf` and adds real `$GIT_DIR`/`$ROOT` paths, but `<hook>` is still a literal placeholder string in the output. The fix description said it would show actual hook names, but the format string `%s/hooks/<hook> -> %s/.githooks/<hook>` still emits literal `<hook>` tokens.

**Suggested fix:** Either emit one line per hook inside the apply loop (most informative), or make the placeholder obviously a pattern by using curly braces `{hook}` to signal it's a template token, not the actual hook name.

tests/e2e/setup_test.go
```diff
@@ -0,0 +1,372 @@
...
+			repo2 := setupRepo(t)
+			populateGithooks(t, repo2)
+			row.setup(t, repo2)
+
+			r2 := runSetup(t, repo2, "--force")
+			if r2.ExitCode != 0 {
+				t.Fatalf("with --force: expected exit 0, got %d\nstdout: %s\nstderr: %s",
+					r2.ExitCode, r2.Stdout, r2.Stderr)
+			}
+			assertContains(t, r2.Stderr, "--force")
```
lugassawan, line 275: **Medium** — the fix added `printf 'Unset core.hooksPath (was: %s)\\n' "$existing_hp"` to stdout when `--force` unsets a conflicting `core.hooksPath`. But `TestSetupConflicts` for the `"custom core.hooksPath"` row never asserts that this message appears in `r2.Stdout`. The new output is entirely untested.

**Suggested fix:** After `assertContains(t, r2.Stderr, "--force")`, add:
```go
if row.name == "custom core.hooksPath" {
    assertContains(t, r2.Stdout, "Unset core.hooksPath")
}
```
Or add a `wantUnsetMsg bool` field to `conflictRow` for a more generalisable assertion.

scripts/setup.sh
lugassawan: Addressed in 7763c38: removed the literal `<hook>` placeholder; the message now reads `%s/hooks/ -> %s/.githooks/`, which is accurate without looking like an unexpanded template.

tests/e2e/setup_test.go
lugassawan: Addressed in 7763c38: added `forceStdout string` field to `conflictRow` (scoped per-row, avoids false positives on the other two conflict rows) and the corresponding `assertContains(t, r2.Stdout, row.forceStdout)` assertion.

PR #204, opened 2026-05-21
[chore] Extract update orchestration into internal/updater.Runner

lugassawan:
## Summary
- Move the ~138-line `cmd/update.go` `RunE` closure into `internal/updater.Runner` with function-valued struct fields as injectable seams
- Each of the 13 `errhint.WithFix` error surfaces now has a focused unit test that injects a fake at exactly one stage — previously untestable because CI can't trigger a real update
- `cmd/update.go` `RunE` shrinks from ~138 to ~17 LOC (thin wrapper: dev-build guard + spinner + `NewRunner` + `Run`)
- `cmd/root.go` `Execute()` gains `signal.NotifyContext` so Ctrl-C cancels a mid-update
- All hint strings preserved verbatim (including em-dash U+2014 and escaped PATH export quotes)

## Test Plan

## Issue

Closes #195

### DIFF
--- a/internal/updater/runner.go
+++ b/internal/updater/runner.go
@@ -1,0 +1,244 @@
+package updater
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"os"
+	"os/exec"
+	"path/filepath"
+	"strings"
+
+	"github.com/lugassawan/rimba/internal/errhint"
+	"github.com/lugassawan/rimba/internal/spinner"
+)
+
+const (
+	retryUpdateHint = "retry: rimba update"
+	reinstallHint   = "reinstall rimba: https://github.com/lugassawan/rimba#installation"
+)
+
+// Runner orchestrates the self-update flow with injectable seams for testability.
+// Each unexported field maps to one stage of the update pipeline; NewRunner wires
+// production defaults. ctx is checked once before the destructive install stage —
+// os.Rename in Replace is atomic so a check inside Replace is pointless.
+// Post-Replace pre-verify crash: a SIGINT/crash after Replace but before verify
+// leaves a working new binary; recovery is `rimba version`.
+type Runner struct {
+	Version   string
+	Out       io.Writer
+	Spinner   *spinner.Spinner
+	OnSuccess func()
+
+	check          func() (*CheckResult, error)
+	download       func(url string) (string, error)
+	prepareBinary  func(path string) error
+	executable     func() (string, error)
+	evalSymlinks   func(path string) (string, error)
+	replace        func(currentBinary, newBinary string) error
+	userInstallDir func() (string, error)
+	mkdirAll       func(path string, perm os.FileMode) error
+	stat           func(path string) (os.FileInfo, error)
+	readFile       func(path string) ([]byte, error)
+	writeFile      func(path string, data []byte, perm os.FileMode) error
+	ensurePath     func(dir string) error
+	execCommand    func(ctx context.Context, name string, args ...string) ([]byte, error)
+	cleanupTempDir func(binaryPath string)
+}
+
+// NewRunner creates a Runner with production defaults wired.
+func NewRunner(version string) *Runner {
+	u := New(version)
+	r := &Runner{
+		Version: version,
+		Out:     os.Stdout,
+		Spinner: spinner.New(spinner.Options{Writer: io.Discard}),
+	}
+	r.check = u.Check
+	r.download = u.Download
+	r.prepareBinary = PrepareBinary
+	r.executable = os.Executable
+	r.evalSymlinks = filepath.EvalSymlinks
+	r.replace = Replace
+	r.userInstallDir = UserInstallDir
+	r.mkdirAll = os.MkdirAll
+	r.stat = os.Stat
+	r.readFile = os.ReadFile
+	r.writeFile = os.WriteFile
+	r.ensurePath = EnsurePath
+	r.execCommand = defaultExecCommand
+	r.cleanupTempDir = CleanupTempDir
+	return r
+}
+
+func defaultExecCommand(ctx context.Context, name string, args ...string) ([]byte, error) {
+	return exec.CommandContext(ctx, filepath.Clean(name), args...).Output() //nolint:gosec // path comes from os.Executable
+}
+
+// Run executes the full update pipeline: check → download → prepare → locate →
+// install → verify. ctx is checked before the destructive install stage.
+func (r *Runner) Run(ctx context.Context) error {
+	r.Spinner.Start("Checking for updates...")
+	result, err := r.check()
+	if err != nil {
+		return errhint.WithFix(
+			fmt.Errorf("checking for updates: %w", err),
+			"check network connectivity, or set GITHUB_TOKEN if rate limited",
+		)
+	}
+
+	if result.UpToDate {
+		r.Spinner.Stop()
+		fmt.Fprintf(r.Out, "Already up to date (%s).\n", result.CurrentVersion)
+		return nil
+	}
+
+	r.Spinner.Stop()
+	fmt.Fprintf(r.Out, "New version available: %s → %s\n", result.CurrentVersion, result.LatestVersion)
+
+	r.Spinner.Start("Downloading...")
+	newBinary, err := r.download(result.DownloadURL)
+	if err != nil {
+		return errhint.WithFix(
+			fmt.Errorf("downloading update: %w", err),
+			"check network connectivity and retry: rimba update",
+		)
+	}
+	defer r.cleanupTempDir(newBinary)
+
+	if err := r.prepareBinary(newBinary); err != nil {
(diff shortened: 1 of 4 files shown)


internal/updater/runner.go
```diff
@@ -0,0 +1,244 @@
...
+func defaultExecCommand(ctx context.Context, name string, args ...string) ([]byte, error) {
+	return exec.CommandContext(ctx, filepath.Clean(name), args...).Output() //nolint:gosec // path comes from os.Executable
+}
+
+// Run executes the full update pipeline: check → download → prepare → locate →
+// install → verify. ctx is checked before the destructive install stage.
+func (r *Runner) Run(ctx context.Context) error {
+	r.Spinner.Start("Checking for updates...")
+	result, err := r.check()
+	if err != nil {
```
lugassawan, line 89: **High — nil-pointer panic if `Spinner` or `Out` is nil**

`Runner` is an exported struct with exported fields. Any caller who zero-initializes `Runner{}` or sets `r.Spinner = nil` after construction will panic on the first `r.Spinner.Start(...)` call. `NewRunner` always wires a no-op Spinner but that invariant is invisible to callers.

Suggested guard at the top of `Run`:

```go
func (r *Runner) Run(ctx context.Context) error {
    if r.Spinner == nil {
        r.Spinner = spinner.New(spinner.Options{Writer: io.Discard})
    }
    if r.Out == nil {
        r.Out = io.Discard
    }
```

internal/updater/runner.go
```diff
@@ -0,0 +1,244 @@
...
+type Runner struct {
+	Version   string
+	Out       io.Writer
+	Spinner   *spinner.Spinner
+	OnSuccess func()
+
+	check          func() (*CheckResult, error)
+	download       func(url string) (string, error)
+	prepareBinary  func(path string) error
+	executable     func() (string, error)
```
lugassawan, line 36: **Medium — `check` and `download` seams don't accept `context.Context`**

A SIGINT during the GitHub API call or binary download cannot cancel either operation — the process must wait for the network timeout even after the signal is received. The `ctx` passed to `Run` is live and wired from `signal.NotifyContext` in `root.go`.

Two options:
1. Change seam signatures to `func(context.Context, ...)` and thread ctx through.
2. Document the limitation in the struct comment if deferring is intentional.

internal/updater/runner.go
```diff
@@ -0,0 +1,244 @@
...
+
+	currentBinary, err := r.locateCurrentBinary()
+	if err != nil {
+		return err
+	}
+
+	if ctx.Err() != nil {
+		return ctx.Err()
+	}
+
```
lugassawan, line 127: **Medium — `ctx.Err()` returned unwrapped, breaks hint contract**

Every other early-exit in `Run` returns an `errhint.WithFix` error so the CLI layer formats a `To fix:` footer. Returning raw `ctx.Err()` (which is `context.Canceled` or `context.DeadlineExceeded`) means the user sees a bare `context canceled` with no recovery hint.

Suggested fix:
```go
if ctx.Err() != nil {
    return errhint.WithFix(ctx.Err(), retryUpdateHint)
}
```

internal/updater/runner_test.go
```diff
@@ -0,0 +1,334 @@
...
+		},
+		{
+			name: "execCommand fails",
+			setupFn: func(_ *testing.T, r *Runner) {
+				r.execCommand = func(ctx context.Context, name string, args ...string) ([]byte, error) {
+					return nil, errors.New("exec failed")
+				}
+			},
+			wantPrefix: "verifying new binary:",
+			wantHint: "the new binary at " + testCurrentBinary +
```
lugassawan, line 223: **Medium — `execCommand` failure test only covers the direct-install path**

The `wantHint` hard-codes `testCurrentBinary` as the installed path, but if the permission-fallback ran first, `installedBinary` would be `testUserDir + "/rimba"`. The table entry does not override `replace` to fail, so `installedBinary == testCurrentBinary` — the assertion passes but the fallback-install verify path is untested.

Suggested: add a second sub-case (or standalone test) where `replace` returns a permission error first, then `execCommand` fails, and assert the hint mentions `testUserDir + "/rimba"`.

internal/updater/runner.go
```diff
@@ -0,0 +1,244 @@
...
+		}
+		return r.installToUserDir(currentBinary, newBinary)
+	}
+	return currentBinary, nil
+}
+
+func (r *Runner) installToUserDir(currentBinary, newBinary string) (string, error) {
+	r.Spinner.Stop()
+	userDir, err := r.userInstallDir()
+	if err != nil {
```
lugassawan, line 184: **Low — `installToUserDir` calls `r.Spinner.Stop()` at entry, but the outer `Run` also stops the spinner after install**

The spinner is already stopped at the outer level after `r.install` returns. `installToUserDir` (called from inside `r.install`) also calls `r.Spinner.Stop()` unconditionally. `Spinner.Stop` appears idempotent (via the `running` flag + mutex), so no functional bug — but the dual-stop is confusing and couples `installToUserDir` to spinner lifecycle management it doesn't need to own.

Option: remove the `r.Spinner.Stop()` call here and let the outer `Run` own the stop-after-install lifecycle exclusively.

internal/updater/runner_test.go
```diff
@@ -0,0 +1,334 @@
...
+)
+
+// newTestRunner returns a Runner wired with fake seams that all succeed, plus a
+// buffer capturing stdout. Tests override individual seams to exercise one failure path.
+func newTestRunner(t *testing.T) (*Runner, *bytes.Buffer) {
+	t.Helper()
+	var buf bytes.Buffer
+	r := NewRunner("1.0.0")
+	r.Out = &buf
+	r.check = func() (*CheckResult, error) {
```
lugassawan: **Low — `newTestRunner` calls `NewRunner` unnecessarily**

`NewRunner` allocates a real `*Updater` (via `New(version)`) and a `spinner.New(...)` — both immediately overwritten. This couples test setup to production init code and adds minor overhead per test call.

Option: construct directly:
```go
r := &Runner{
    Version: "1.0.0",
    Out:     &buf,
    Spinner: spinner.New(spinner.Options{Writer: io.Discard}),
}
```
Then wire only the seams under test. Makes the test fully self-contained.

internal/updater/runner.go
```diff
@@ -0,0 +1,244 @@
...
+	r.readFile = os.ReadFile
+	r.writeFile = os.WriteFile
+	r.ensurePath = EnsurePath
+	r.execCommand = defaultExecCommand
+	r.cleanupTempDir = CleanupTempDir
+	return r
+}
+
+func defaultExecCommand(ctx context.Context, name string, args ...string) ([]byte, error) {
+	return exec.CommandContext(ctx, filepath.Clean(name), args...).Output() //nolint:gosec // path comes from os.Executable
```
lugassawan: **Low — `nolint:gosec` comment could be more precise**

The comment says `// path comes from os.Executable` but `name` here is actually `installedBinary` (the just-written binary path), not the result of `os.Executable()` directly. More accurate:
```go
//nolint:gosec // G204: name is installedBinary, written by this process
```
Pure clarity — no functional impact.

internal/updater/runner.go
lugassawan: Addressed in 222b92b: Added nil guards for `Spinner` and `Out` at the top of `Run` — both are initialized to safe defaults when nil.

internal/updater/runner.go
lugassawan: Addressed in 222b92b: Changed `check` and `download` seam types to accept `context.Context` as first arg; `NewRunner` wraps `u.Check`/`u.Download` in context-forwarding closures.

internal/updater/runner.go
lugassawan: Addressed in 222b92b: `ctx.Err()` now wrapped with `errhint.WithFix(err, retryUpdateHint)` to maintain the hint contract.

internal/updater/runner_test.go
lugassawan: Addressed in 222b92b: Added `TestRunVerifyFailsAfterFallbackInstall` — covers the perm-fallback path where `execCommand` fails and asserts the hint references the user-dir binary path.

internal/updater/runner.go
lugassawan: Addressed in 222b92b: Removed the redundant `r.Spinner.Stop()` from `installToUserDir` — the outer `Run` already stops the spinner before calling `install`.

internal/updater/runner_test.go
lugassawan: Addressed in 222b92b: `newTestRunner` now uses a `&Runner{...}` struct literal; added `TestNewRunner` as a dedicated smoke test for `NewRunner` (keeping total coverage ≥97%).

internal/updater/runner.go
lugassawan: Addressed in 222b92b: Updated the comment to `// G204: name is installedBinary, written by this process`.

PR #212, opened 2026-05-25
[chore] Audit hardening: security + reliability fixes

lugassawan:
## Summary

- **Unit A** — HTTP client 30s timeout + context threading in `Updater.Check`/`Download`; cancellable downloads and rate-limit-aware error hints
- **Unit B** — tar zip-bomb guard in `extractBinary`: skip non-regular-file entries (`TypeReg` check) and cap decompressed output at 100 MiB via `io.LimitReader`
- **Unit C** — `EnsurePath` false-positive fix: skip comment lines (`# ...`) when scanning for existing PATH exports
- **Unit D** — `HeadRefName` (branch name) validation in `FetchPRMeta`: permissive regex (`^[a-zA-Z0-9._/-]+$`), rejects leading `-`/`..`/`/` to prevent option injection
- **Unit E** — Ctrl-C context propagation: `cmd/status.go` uses `cmd.Context()` (signal-aware) instead of `context.Background()`; `StatusDashboard` adds entry-guard `ctx.Err()` check
- **Unit F** — Clone error aggregation in `cloneRecursive`: collects partial failures via `errors.Join` instead of silently discarding them
- **Unit G** — `govulncheck` CI job (informational `|| true` until go toolchain is bumped past current stdlib CVEs — see TODO in ci.yml)

## Test Plan

- [x] Unit tests pass (`make test`) — 97.5% coverage (≥97% gate)
- [x] E2E tests pass (`make test-e2e`) — 276/276
- [x] `govulncheck ./...` run locally; findings are go1.26.0 stdlib CVEs fixed in go1.26.1 (CI uses `go-version: stable`)

### Type-tailored checks ([chore]/security)

- [ ] `rimba update` still completes a real self-update end-to-end
- [ ] `rimba status` respects Ctrl-C (sends SIGINT mid-run, confirms clean exit)
- [ ] `rimba add <pr>` with a branch name containing `/` (e.g. `dependabot/...`) is accepted
- [ ] `rimba add <pr>` with a branch name starting with `-` or containing `..` is rejected

## Issue

Closes #206

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,6 +55,9 @@ jobs:
 
   vuln:
     runs-on: ubuntu-latest
+    # TODO: remove continue-on-error once go.mod toolchain is bumped to ≥go1.26.1
+    # to resolve stdlib CVEs (GO-2026-4599 etc.) and gate is confirmed clean.
+    continue-on-error: true
     steps:
       - uses: actions/checkout@v6
       - uses: actions/setup-go@v6
@@ -62,7 +65,4 @@ jobs:
           go-version: stable
           cache: true
       - run: go install golang.org/x/vuln/cmd/govulncheck@latest
-        # TODO: remove || true once the local toolchain (go1.26.0) is updated to ≥1.26.1
-        # and all stdlib CVEs (GO-2026-4599 etc.) are confirmed resolved.
-        # Tracked: upgrade go directive in go.mod and CI toolchain, then make this gate blocking.
-      - run: govulncheck ./... || true
+      - run: govulncheck ./...
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -52,3 +52,17 @@ jobs:
           go-version: stable
           cache: true
       - run: make build
+
+  vuln:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6
+      - uses: actions/setup-go@v6
+        with:
+          go-version: stable
+          cache: true
+      - run: go install golang.org/x/vuln/cmd/govulncheck@latest
+        # TODO: remove || true once the local toolchain (go1.26.0) is updated to ≥1.26.1
+        # and all stdlib CVEs (GO-2026-4599 etc.) are confirmed resolved.
+        # Tracked: upgrade go directive in go.mod and CI toolchain, then make this gate blocking.
+      - run: govulncheck ./... || true
--- a/internal/deps/clone.go
+++ b/internal/deps/clone.go
@@ -63,7 +63,7 @@ func cloneRecursive(srcWT, dstWT string, mod Module) error {
 	_ = filepath.WalkDir(searchRoot, walkCloneFunc(srcWT, dstWT, baseName, &cloneErrs))
 
 	if err := cloneExtraDirs(srcWT, dstWT, mod.ExtraDirs); err != nil {
-		return err
+		cloneErrs = append(cloneErrs, err)
 	}
 
 	return errors.Join(cloneErrs...)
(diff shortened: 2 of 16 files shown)


.github/workflows/ci.yml
```diff
@@ -52,3 +52,17 @@ jobs:
...
+      - uses: actions/checkout@v6
+      - uses: actions/setup-go@v6
+        with:
+          go-version: stable
+          cache: true
+      - run: go install golang.org/x/vuln/cmd/govulncheck@latest
+        # TODO: remove || true once the local toolchain (go1.26.0) is updated to ≥1.26.1
+        # and all stdlib CVEs (GO-2026-4599 etc.) are confirmed resolved.
+        # Tracked: upgrade go directive in go.mod and CI toolchain, then make this gate blocking.
+      - run: govulncheck ./... || true
```
lugassawan: **Medium | `.github/workflows/ci.yml:68`**

`govulncheck ./... || true` makes this job unconditionally pass regardless of findings — the security gate is permanently bypassed.

**Why it matters:** The `|| true` will silently suppress all future CVEs discovered after the toolchain is bumped, not just the current go1.26.0 stdlib findings. The TODO has no linked issue number or deadline, so there is no mechanism to make this blocking.

**Suggested fix:** Use `continue-on-error: true` at the job level (makes the intent visible in the GitHub UI) and link a tracking issue in the comment:

```yaml
  vuln:
    runs-on: ubuntu-latest
    continue-on-error: true  # TODO: remove once go.mod toolchain >= go1.26.1 (issue #NNN)
    steps:
      ...
      - run: govulncheck ./...
```

This keeps the informational intent but makes it auditable — a yellow job icon signals "there are findings" rather than a silent green.

internal/deps/clone.go
```diff
@@ -56,20 +58,24 @@ func cloneRecursive(srcWT, dstWT string, mod Module) error {
...
 	}
 
+	var cloneErrs []error
 	baseName := filepath.Base(mod.Dir)
-	err := filepath.WalkDir(searchRoot, walkCloneFunc(srcWT, dstWT, baseName))
-	if err != nil {
+	_ = filepath.WalkDir(searchRoot, walkCloneFunc(srcWT, dstWT, baseName, &cloneErrs))
+
+	if err := cloneExtraDirs(srcWT, dstWT, mod.ExtraDirs); err != nil {
 		return err
```
lugassawan: **Medium | `internal/deps/clone.go:66`**

Early `return err` from `cloneExtraDirs` silently discards all walk errors collected in `cloneErrs`.

**Why it matters:** The error-aggregation refactor was specifically meant to surface partial failures. When `cloneExtraDirs` fails, the walk failures are lost entirely — callers see only the extra-dir error and have no visibility into which module subdirectories failed to clone.

**Suggested fix:**

```go
if err := cloneExtraDirs(srcWT, dstWT, mod.ExtraDirs); err != nil {
    cloneErrs = append(cloneErrs, err)
}
return errors.Join(cloneErrs...)
```

The existing `TestCloneModuleRecursiveExtraDirError` does not exercise a simultaneous walk failure, so this path is untested.

.github/workflows/ci.yml
lugassawan: Addressed in 508a4a4: changed from `|| true` step-level pass to `continue-on-error: true` at job level, so the job shows as yellow/neutral in the GitHub UI when govulncheck finds CVEs instead of silently passing green.

internal/deps/clone.go
lugassawan: Addressed in 508a4a4: changed `return err` to `cloneErrs = append(cloneErrs, err)` in `cloneRecursive`, so extra-dir failures are now aggregated into the walk-error slice and surfaced via `errors.Join` rather than causing an early return that discards all prior walk errors.

PR #213, opened 2026-05-27
[fix] promote flow: surface stash apply/drop errors instead of masking them

lugassawan:
## Summary
- `applyStashToWorktree` previously labelled every `StashApply` failure as a conflict, masking the real cause (e.g. "local changes would be overwritten")
- Drop errors after a successful apply were silently discarded with `_ = git.StashDrop(...)`, leaving orphaned stash entries with no warning
- Fix: distinguish real `CONFLICT` failures (conflict-specific hint + SHA) from other apply failures (surface true cause, preserve stash), and surface drop errors rather than swallowing them

## Test Plan
- [x] Unit tests pass (`make test`) — including 2 new tests: `TestPromoteBranchStashApplyNonConflictFails`, `TestPromoteBranchStashDropFails`
- [x] Coverage ≥ 97% (`make test-coverage`) — 97.5%

## Issue

Closes #207

### DIFF
--- a/internal/operations/add_branch_test.go
+++ b/internal/operations/add_branch_test.go
@@ -393,17 +393,25 @@ func conflictRunFn(porcelain, wtDir string) func(args ...string) (string, error)
 	}
 }
 
-// stashOutcomeDispatch handles stash-operation args for promote mocks, injecting
-// the given apply/drop errors. Extracted so its cyclomatic complexity does not
-// count toward the closure in stashOutcomeRunInDirFn.
-func stashOutcomeDispatch(stashDropped *bool, applyErr, dropErr error, args []string) (string, error, bool) {
+// stashOutcomeDispatchA handles push / rev-parse / switch for promote mocks.
+func stashOutcomeDispatchA(args []string) (string, bool) {
 	switch {
 	case len(args) >= 2 && args[0] == gitCmdStash && args[1] == gitSubcmdPush:
-		return "", nil, true
+		return "", true
 	case len(args) >= 2 && args[0] == cmdRevParse && args[1] == "stash@{0}":
-		return stashSHATest, nil, true
+		return stashSHATest, true
 	case len(args) >= 2 && args[0] == gitCmdSwitch:
-		return "", nil, true
+		return "", true
+	}
+	return "", false
+}
+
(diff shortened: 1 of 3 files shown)


internal/operations/add_branch_test.go
```diff
@@ -447,3 +450,56 @@ func TestPromoteBranchStashApplyConflict(t *testing.T) {
...
+	}
+	if strings.Contains(err.Error(), "had conflicts") {
+		t.Errorf("error %q should NOT mention 'had conflicts' for a non-conflict failure", err)
+	}
+	if stashDropped {
+		t.Error("stash should NOT be dropped on apply failure")
+	}
+}
+
+func TestPromoteBranchStashDropFails(t *testing.T) {
```
lugassawan, line 501: **High | `internal/operations/add_branch_test.go:484–505` | `dropErr` injection is unreachable — test exercises stash-list failure, not drop failure**

`stashOutcomeDispatch` has no arm for `stash list`, which `git.StashDrop` calls internally via `stashRefBySHA` before issuing the actual drop. The list call falls through returning `"", nil, false`, so `stashRefBySHA` fails with "stash entry not found" — `errGitFailed` is never injected and `stashDropped` stays `false`. The test passes because the error wrapping always contains `"could not drop"`, but it is testing a stash-list-failure path, not the intended drop-failure path.

Add a `stash list` arm to `stashOutcomeDispatch` using the existing `stashListLine` constant (already used in the happy-path helper `promoteRunInDirStashB`):

```go
case len(args) >= 3 && args[0] == gitCmdStash && args[1] == gitSubcmdList:
    return stashListLine, nil, true
case len(args) >= 2 && args[0] == gitCmdStash && args[1] == gitSubcmdDrop:
    *stashDropped = true
    return "", dropErr, true
```

Also consider asserting `stashDropped == true` in `TestPromoteBranchStashDropFails` to confirm the drop command was actually reached.

lugassawan: Addressed in a7f26feb65c6e696d2aca8551835305cccf16804: split stashOutcomeDispatch into A/B helpers, added stashListLine arm so stashRefBySHA succeeds and the actual stash drop command is reached, and added stashDropped assertion to confirm the drop path is exercised.

PR #214, opened 2026-05-27
[fix] unify partial-failure recovery and auto-abort failed merges

lugassawan:
## Summary

- Unified the "worktree removed but branch delete failed" recovery error across `remove`, `clean`, and `merge` — previously each cmd re-templated its own hint string; now `branchDeleteFailedErr` in `operations.go` is the single source of truth and cmd layer prints the stored error
- Added `MergeInProgress` and `MergeAbort` helpers to `internal/git`
- `MergeWorktree` now auto-aborts a failed merge via `git merge --abort` when `MERGE_HEAD` is present, restoring the target worktree to clean state; if the abort itself fails, both errors are surfaced with a manual-cleanup hint; if no merge was in progress (fast-forward pre-check failure, bad ref), reports "target unchanged" instead of the misleading "restored" message
- Cmd layer (`cmd/clean.go`, `cmd/merge.go`) changed from re-templating to `fmt.Fprintln(cmd.OutOrStdout(), item.Error)` — thin presenter pattern

## Test Plan

- [x] Coverage ≥ 97% confirmed (97.5%)

### Type-tailored checks ([fix])

- [x] `TestMergeInProgress` — false on clean repo, true after conflicting merge
- [x] `TestMergeAbort` — abort restores clean state and clears MERGE_HEAD
- [x] `TestMergeWorktreeMergeConflict` — error contains "merge failed" + "restored to pre-merge state"
- [x] `TestMergeWorktreeMergeFailsAbortAlsoFails` — error contains "rollback failed"
- [x] `TestMergeWorktreeMergeFailsNoMergeInProgress` — error contains "target … unchanged"
- [x] `TestMergeWorktreeCleanupBranchDeleteFails` — `result.RemoveError` contains unified hint
- [x] `TestRemoveAndCleanupBranchDeleteFails` — error contains "failed to delete branch" + "git branch -D"
- [x] `TestPrintCleanedItemsAllBranches` — clean cmd prints stored error verbatim

## Issue

Closes #208

### DIFF
--- a/internal/git/merge.go
+++ b/internal/git/merge.go
@@ -10,3 +10,15 @@ func Merge(r Runner, dir, branch string, noFF bool) error {
 	_, err := r.RunInDir(dir, args...)
 	return err
 }
+
+// MergeInProgress reports whether a merge is in progress in dir (MERGE_HEAD exists).
+func MergeInProgress(r Runner, dir string) bool {
+	_, err := r.RunInDir(dir, "rev-parse", "--verify", "-q", "MERGE_HEAD")
+	return err == nil
+}
+
+// MergeAbort runs `git merge --abort` in dir.
+func MergeAbort(r Runner, dir string) error {
+	_, err := r.RunInDir(dir, "merge", "--abort")
+	return err
+}
--- a/internal/git/merge.go
+++ b/internal/git/merge.go
@@ -1,5 +1,10 @@
 package git
 
+import (
+	"fmt"
+	"strings"
+)
+
 // Merge runs `git merge [--no-ff] <branch>` inside the given directory.
 func Merge(r Runner, dir, branch string, noFF bool) error {
 	args := []string{"merge"}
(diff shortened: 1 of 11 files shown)


internal/git/merge.go
```diff
@@ -10,3 +10,15 @@ func Merge(r Runner, dir, branch string, noFF bool) error {
 	_, err := r.RunInDir(dir, args...)
 	return err
 }
+
+// MergeInProgress reports whether a merge is in progress in dir (MERGE_HEAD exists).
+func MergeInProgress(r Runner, dir string) bool {
```
lugassawan: **Medium: `MergeInProgress` returns `bool`, silently treating infrastructure errors as "no merge in progress"**

```go
func MergeInProgress(r Runner, dir string) bool {
    _, err := r.RunInDir(dir, "rev-parse", "--verify", "-q", "MERGE_HEAD")
    return err == nil
}
```

If `RunInDir` fails for a reason other than MERGE_HEAD not existing (disk I/O error, git binary missing, bad directory), `MergeInProgress` returns `false` and `abortFailedMerge` reports "target unchanged" — a misleading hint when a merge may actually be in progress.

**Suggested fix:** Return `(bool, error)` from `MergeInProgress`. In `abortFailedMerge`, if the error is non-nil, fall through to the manual-cleanup hint (`cd $dir && git merge --abort`) rather than assuming no merge is in progress. This makes the API self-documenting and avoids silently misleading users in broken environments.

lugassawan: Addressed in 9d8fcdf41b4e2529dc1a3a03610e26b6bd707074: changed `MergeInProgress` to return `(bool, error)`. The error is surfaced for non-ref-not-found failures; `abortFailedMerge` now emits a conservative manual-cleanup hint when merge state cannot be determined.

PR #216, opened 2026-05-27
[feat] Prune stale remote-tracking refs during rimba clean

lugassawan:
## Summary
- `rimba clean` (default mode) now runs `git remote prune origin` after worktree prune, removing stale `refs/remotes/origin/*` tracking refs that linger after PR branches are deleted on the remote
- `--dry-run` previews remote-ref pruning without deleting; output clearly distinguishes worktree-ref pruning from remote-ref pruning (two separate lines)
- No `origin` remote → skipped gracefully with a non-fatal notice (`No 'origin' remote; skipped remote-ref prune.`), not an error
- MCP `clean` tool (mode=prune) gains a `remote_pruned []string` field (omitempty) in its JSON result, keeping parity with the CLI default

## Test Plan

## Issue

Closes #211

### DIFF
--- a/cmd/clean.go
+++ b/cmd/clean.go
@@ -93,11 +93,13 @@ func cleanPrune(cmd *cobra.Command, r git.Runner) error {
 // cleanRemotePrune prunes stale origin remote-tracking refs, skipping gracefully
 // when there is no origin remote.
 func cleanRemotePrune(cmd *cobra.Command, r git.Runner, s *spinner.Spinner, dryRun bool) error {
+	s.Start("Pruning remote-tracking refs...")
 	if !git.RemoteExists(r, "origin") {
+		s.Stop()
 		fmt.Fprintln(cmd.OutOrStdout(), "No 'origin' remote; skipped remote-ref prune.")
 		return nil
 	}
-	s.Start("Pruning remote-tracking refs...")
+	s.Update("Pruning remote-tracking refs...")
 	refs, err := git.RemotePrune(r, "origin", dryRun)
 	if err != nil {
 		return err
--- a/cmd/clean.go
+++ b/cmd/clean.go
@@ -26,8 +26,9 @@ const (
 var cleanCmd = &cobra.Command{
 	Use:   "clean",
 	Short: "Prune stale worktree references or remove merged worktrees",
-	Long:  "Runs git worktree prune to clean up stale references. Use --merged to detect and remove worktrees whose branches have been merged into main.",
+	Long:  "Runs git worktree prune to clean up stale references and prunes stale remote-tracking refs from origin. Use --merged to detect and remove worktrees whose branches have been merged into main.",
 	Example: `  rimba clean
+  rimba clean --dry-run
   rimba clean --merged
   rimba clean --stale --stale-days 7`,
 	Annotations: map[string]string{"skipConfig": "true"},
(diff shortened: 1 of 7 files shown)


internal/git/remote.go
```diff
@@ -1,13 +1,51 @@
...
 }
+
+// parsePrunedRefs extracts ref names from `git remote prune` output lines like
+// ` * [pruned] origin/x` (live) and ` * [would prune] origin/x` (dry-run).
+func parsePrunedRefs(out string) []string {
+	refs := []string{}
+	for line := range strings.SplitSeq(out, "\n") {
+		if strings.Contains(line, "[pruned]") || strings.Contains(line, "[would prune]") {
+			fields := strings.Fields(line)
+			refs = append(refs, fields[len(fields)-1])
```
lugassawan: **High** — Silent data corruption when a  line has no trailing ref name.

`parsePrunedRefs` checks `strings.Contains(line, "[pruned]")` but then unconditionally indexes `fields[len(fields)-1]`. A line like `   [pruned]   ` (tag with only surrounding whitespace, which cannot reach here in practice, but guards against future git format changes) would have `fields == ["[pruned]"]` and return the tag itself as the ref name — silent wrong output rather than a panic.

**Suggested fix:**
```go
fields := strings.Fields(line)
if len(fields) >= 2 {
    refs = append(refs, fields[len(fields)-1])
}
```

This matches the robustness pattern used in other git-output parsers in this codebase (e.g. `MergedBranches`) and defends against unexpected git locale output.

cmd/clean.go
```diff
@@ -86,6 +87,31 @@ func cleanPrune(cmd *cobra.Command, r git.Runner) error {
 		fmt.Fprintln(cmd.OutOrStdout(), "Pruned stale worktree references.")
 	}
 
+	return cleanRemotePrune(cmd, r, s, dryRun)
+}
+
+// cleanRemotePrune prunes stale origin remote-tracking refs, skipping gracefully
+// when there is no origin remote.
+func cleanRemotePrune(cmd *cobra.Command, r git.Runner, s *spinner.Spinner, dryRun bool) error {
+	if !git.RemoteExists(r, "origin") {
```
lugassawan, line 97: **Medium** — Spinner is started *after* the `RemoteExists` check, leaving the UI frozen during the `git remote get-url` subprocess.

The worktree-prune block above (and `cleanFetchMergeRef`) starts the spinner *before* the blocking git call. `cleanRemotePrune` diverges from this pattern: it calls `git.RemoteExists` (which runs a subprocess) before calling `s.Start`.

**Suggested fix:** start the spinner at the top of `cleanRemotePrune` and update the message on the no-origin early-return path, mirroring the existing pattern:

```go
func cleanRemotePrune(cmd *cobra.Command, r git.Runner, s *spinner.Spinner, dryRun bool) error {
    s.Start("Checking remote-tracking refs...")
    if !git.RemoteExists(r, "origin") {
        s.Stop()
        fmt.Fprintln(cmd.OutOrStdout(), "No 'origin' remote; skipped remote-ref prune.")
        return nil
    }
    s.Update("Pruning remote-tracking refs...")
    // ...
```

internal/git/remote.go
lugassawan: Addressed in 4b28f45f234b70cd581e0427ff7380b5b536cc63: added `if fields := strings.Fields(line); len(fields) >= 2` guard before indexing `fields[len(fields)-1]`.

cmd/clean.go
lugassawan: Addressed in 4b28f45f234b70cd581e0427ff7380b5b536cc63: spinner now starts with `s.Start()` before the `RemoteExists` call, then `s.Update()` is used for the prune step, matching the project's established spinner pattern.

PR #217, opened 2026-05-27
[docs] Sync README and command reference with current CLI behavior

lugassawan:
## Summary

- Fix stale prefix wording: "configurable prefix (e.g. `feat/`, `fix/`)" → fixed set of 6 (`feature/`, `bugfix/`, `hotfix/`, `docs/`, `test/`, `chore/`) selected via flags
- Correct `copy_files` defaults: add `.env.local`, demote `.vscode/` from default to illustrative example
- Fix `rimba init` tier flags: `--user`/`--team`/`--local` → `--agents`, `--agents --local`, `-g`/`--global`
- Document `rimba add branch:<branch>` promote mode with constraints (must be checked-out branch, not default; stash transfers dirty state; `--source` invalid)
- Add `--dry-run` to flag tables for `remove`, `duplicate`, `archive`, `merge`, `sync`
- Add `--skip-deps` and `--skip-hooks` to `rimba rename` flag table
- Replace single-line version example with real 6-line output format (`rimba v<X>` / `commit:` / `built:` / `os:` / `arch:` / `go:`)
- Reword `cmd/flags.go` error hint away from nonexistent `.rimba/settings.toml [prefixes]` config section

## Test Plan

- [x] Build succeeds (`make build`)
- [x] Every documented flag verified against live `--help` output for all six affected commands
- [x] `./bin/rimba version` output matches the new docs example exactly
- [x] Error hint (`rimba list --type bogus`) no longer references `[prefixes]` config section
- [x] `git grep` confirms no stale strings remain (`prefixes] section`, `--user\`/\`--team`, `feat/\`, \`fix/`)

## Issue

Issue: N/A

### DIFF
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -78,7 +78,7 @@ rimba add branch:feature/my-feature  # Promote current branch to its own worktre
 
 > **PR mode:** `pr:<num>` requires `gh` to be installed and authenticated. For cross-fork PRs, rimba adds a `gh-fork-<owner>` remote automatically. Without `--task`, the task name is derived as `review/<num>-<slug>`.
 
-> **Branch mode:** `branch:<branch>` requires that `<branch>` is the currently checked-out branch in the main repo and is not the default branch. Any uncommitted changes are transferred to the new worktree via `git stash`. `--source` is not valid in `branch:` mode.
+> **Branch mode:** `branch:<branch>` requires that `<branch>` is the currently checked-out branch in the main repo and is not the default branch. Any uncommitted changes are transferred to the new worktree via `git stash`. `--source` is not valid in `branch:` mode. `--skip-deps` and `--skip-hooks` are accepted by the parser but have no effect in `branch:` mode; promote does not run dependency installation or post-create hooks.
 
 | Flag | Description |
 |------|-------------|
@@ -610,7 +610,7 @@ commit: <sha>
 built:  <iso8601>
 os:     <linux|darwin|windows>
 arch:   <amd64|arm64>
-go:     go<X.Y.Z>
+go:     go<version>
 ```
 
 ---
--- a/docs/commands.md
+++ b/docs/commands.md
@@ -61,7 +61,7 @@ rimba init --agents --uninstall   # Remove project-team agent files and MCP regi
 
 ### rimba add
 
-Create a new worktree with a branch named `<prefix><task>` and copy dotfiles from the repo root. In monorepos, prefix the task with a service directory name to create service-scoped branches. Use `pr:<num>` to create a worktree from a GitHub PR's head branch.
+Create a new worktree with a branch named `<prefix><task>` and copy dotfiles from the repo root. In monorepos, prefix the task with a service directory name to create service-scoped branches. Use `pr:<num>` to create a worktree from a GitHub PR's head branch. Use `branch:<branch>` to promote the currently checked-out branch into its own worktree.
 
 ```sh
 rimba add my-feature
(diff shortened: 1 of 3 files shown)


docs/commands.md
```diff
@@ -71,12 +71,15 @@ rimba add auth-api/my-feature        # Monorepo: branch auth-api/feature/my-feat
...
 rimba add pr:123                     # Create worktree from PR #123's head branch
 rimba add pr:123 --task review/auth-tweak  # Override auto-derived task name
+rimba add branch:feature/my-feature  # Promote current branch to its own worktree
 ```
 
 > **Monorepo:** If the first segment before `/` matches a directory in the repo root, rimba treats it as a service scope. The branch uses a 3-segment pattern: `<service>/<prefix>/<task>`. No configuration needed — detection is automatic.
 
 > **PR mode:** `pr:<num>` requires `gh` to be installed and authenticated. For cross-fork PRs, rimba adds a `gh-fork-<owner>` remote automatically. Without `--task`, the task name is derived as `review/<num>-<slug>`.
 
+> **Branch mode:** `branch:<branch>` requires that `<branch>` is the currently checked-out branch in the main repo and is not the default branch. Any uncommitted changes are transferred to the new worktree via `git stash`. `--source` is not valid in `branch:` mode.
```
lugassawan: **Medium** — `branch:` mode omits that `--skip-deps`/`--skip-hooks` are accepted but ignored

`runAddBranch` in `cmd/add.go` calls `operations.PromoteBranch` with no post-create options, so `--skip-deps` and `--skip-hooks` have no effect in `branch:` mode even though the parser accepts them silently.

The new Branch mode callout documents `--source` as invalid but is silent about these two flags — a user will pass them expecting something to happen and get no feedback.

**Suggested fix:** extend the note with: "`--skip-deps` and `--skip-hooks` are accepted by the parser but have no effect in `branch:` mode; promote does not run dependency installation or post-create hooks."

docs/commands.md
```diff
@@ -587,11 +598,19 @@ rimba update --force     # Also works on dev builds
...
-# rimba v<X.Y.Z> (commit: <sha>, built: <iso8601>)
+```
+
+```
+rimba v<X.Y.Z>
+commit: <sha>
+built:  <iso8601>
+os:     <linux|darwin|windows>
+arch:   <amd64|arm64>
+go:     go<X.Y.Z>
```
lugassawan: **Low** — `go:` line placeholder `go<X.Y.Z>` could mislead readers

`runtime.Version()` returns a string that already includes the `go` prefix (e.g. `go1.26.0`), so the full output line is `go:     go1.26.0` — not `go:     1.26.0`. The placeholder `go:     go<X.Y.Z>` is technically correct but the `<X.Y.Z>` part suggests only digits follow, obscuring the double-`go` pattern that surprises readers the first time.

**Suggested fix:** change `go<X.Y.Z>` to `go<version>` or a concrete example like `go1.24.0` — consistent with showing the actual runtime string format.

docs/commands.md
lugassawan: Addressed in 0f48b586000cbcfc780c4364fcb5a1445c98653b: extended the Branch mode callout to note that `--skip-deps` and `--skip-hooks` are accepted by the parser but have no effect in `branch:` mode.

docs/commands.md
lugassawan: Addressed in 0f48b586000cbcfc780c4364fcb5a1445c98653b: changed `go<X.Y.Z>` to `go<version>` to make it clear the placeholder represents the full runtime string (which already includes the `go` prefix).

PR #232, opened 2026-05-30
[chore] pin go1.26.3 toolchain, harden vuln gate

lugassawan:
## Summary

- Bump `go.mod` from `go 1.25.7` → `go 1.26.3` to fix four stdlib CVEs reachable through the auto-updater's TLS/cert path: GO-2026-4599, GO-2026-4600, GO-2026-4601, GO-2026-4870
- Replace `go-version: stable` with `go-version-file: go.mod` in all 7 `setup-go` steps (6 in `ci.yml`, 1 in `release.yml`) — `go.mod` becomes the single source of truth for the toolchain version
- Remove `continue-on-error: true` and its TODO comment from the `vuln` job — govulncheck is now a true blocking gate (`No vulnerabilities found` confirmed locally)
- Apply Go 1.26 `modernize/newexpr` lint fixes: `testutil.Ptr` marked `//go:fix inline`, callsites inlined to `new(v)`, unused imports removed

## Test Plan

- [x] Unit tests pass (`make test`) — 1843 tests, 26 packages
- [x] Linter passes (`make lint`) — 0 issues under go1.26.3
- [x] `govulncheck ./...` — `No vulnerabilities found`
- [x] `go build ./...` — clean under go1.26.3 (auto-fetched via GOTOOLCHAIN=auto)

### Type-tailored checks ([chore])

- [ ] No application behaviour changed — only toolchain, CI config, and test-utility cosmetics
- [ ] `go.sum` unchanged — no new third-party dependencies introduced
- [ ] Release workflow unchanged in behaviour — only `go-version` key renamed

## Issue

Closes #221

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,7 @@ jobs:
       - uses: actions/checkout@v6
       - uses: actions/setup-go@v6
         with:
-          go-version: stable
+          go-version-file: go.mod
           cache: true
       - uses: actions/cache@v5
         with:
@@ -29,7 +29,7 @@ jobs:
       - uses: actions/checkout@v6
       - uses: actions/setup-go@v6
         with:
-          go-version: stable
+          go-version-file: go.mod
           cache: true
       - name: Run tests with coverage
         run: make test-coverage
@@ -49,7 +49,7 @@ jobs:
       - uses: actions/checkout@v6
       - uses: actions/setup-go@v6
         with:
-          go-version: stable
+          go-version-file: go.mod
           cache: true
       - run: make build
 
@@ -60,7 +60,7 @@ jobs:
       - uses: actions/checkout@v6
       - uses: actions/setup-go@v6
         with:
-          go-version: stable
+          go-version-file: go.mod
           cache: true
       - run: make test-e2e
 
@@ -71,20 +71,17 @@ jobs:
       - uses: actions/checkout@v6
       - uses: actions/setup-go@v6
         with:
-          go-version: stable
+          go-version-file: go.mod
           cache: true
       - run: make test-race
 
   vuln:
     runs-on: ubuntu-latest
-    # TODO: remove continue-on-error once go.mod toolchain is bumped to ≥go1.26.1
-    # to resolve stdlib CVEs (GO-2026-4599 etc.) and gate is confirmed clean.
-    continue-on-error: true
     steps:
       - uses: actions/checkout@v6
       - uses: actions/setup-go@v6
         with:
-          go-version: stable
+          go-version-file: go.mod
           cache: true
       - run: go install golang.org/x/vuln/cmd/govulncheck@latest
       - run: govulncheck ./...
--- a/go.mod
+++ b/go.mod
@@ -7,6 +7,7 @@ require (
 	github.com/mark3labs/mcp-go v0.54.0
 	github.com/pelletier/go-toml/v2 v2.3.1
 	github.com/spf13/cobra v1.10.2
+	github.com/spf13/pflag v1.0.10
 )
 
 require (
@@ -15,7 +16,6 @@ require (
 	github.com/inconshreveable/mousetrap v1.1.0 // indirect
 	github.com/santhosh-tekuri/jsonschema/v6 v6.0.2 // indirect
 	github.com/spf13/cast v1.10.0 // indirect
-	github.com/spf13/pflag v1.0.10 // indirect
 	github.com/yosida95/uritemplate/v3 v3.0.2 // indirect
 	golang.org/x/text v0.14.0 // indirect
 )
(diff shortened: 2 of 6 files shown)


go.mod
```diff
@@ -1,6 +1,6 @@
 module github.com/lugassawan/rimba
 
-go 1.25.7
+go 1.26.3
```
lugassawan, line 3: **Medium — Missing `toolchain` directive**

The `go 1.26.3` line sets the minimum language version but does not hard-pin the toolchain. When a developer's local or CI environment has a newer toolchain installed (e.g. go1.27.x), `GOTOOLCHAIN=auto` will silently upgrade, drifting from the intended pin. The ticket calls for pinning a concrete toolchain.

**Suggested fix:** add a `toolchain go1.26.3` line immediately after:

```
go 1.26.3
toolchain go1.26.3
```

This is idiomatic Go 1.21+ practice and makes the lock explicit.

testutil/ptr.go
```diff
@@ -1,4 +1,6 @@
 package testutil
 
 // Ptr returns a pointer to v — useful in tests for constructing *T literals inline.
-func Ptr[T any](v T) *T { return &v }
+//
+//go:fix inline
```
lugassawan, line 7: **Medium — `//go:fix inline` on a Go 1.26-only body**

`go fix` will inline all call-sites of `testutil.Ptr` with `new(v)`. This is correct for Go ≥ 1.26, but any caller compiled against an older toolchain would see rewritten code that fails to compile (`new(expr)` is a 1.26 addition). Since this is an internal `testutil` package and `go.mod` now gates at 1.26.3, the risk is contained — but a short comment would prevent future maintainers from being confused by the unusual `new(expr)` form.

**Suggested fix:** add a one-liner above the directive:

```go
// new(v) requires Go ≥ 1.26 (the module minimum).
//go:fix inline
func Ptr[T any](v T) *T { return new(v) }
```

.github/workflows/ci.yml
```diff
@@ -71,20 +71,17 @@ jobs:
...
-    # to resolve stdlib CVEs (GO-2026-4599 etc.) and gate is confirmed clean.
-    continue-on-error: true
     steps:
       - uses: actions/checkout@v6
       - uses: actions/setup-go@v6
         with:
-          go-version: stable
+          go-version-file: go.mod
           cache: true
       - run: go install golang.org/x/vuln/cmd/govulncheck@latest
```
lugassawan, line 86: **Low — `govulncheck@latest` is non-reproducible**

This line is pre-existing (not introduced by this PR), but the stated goal of the PR is to make the vuln gate deterministic. Installing `@latest` means a newly-released govulncheck with stricter checks could break CI on an otherwise clean commit.

**Suggested fix:** pin to a specific version (e.g. `govulncheck@v1.1.4`), or add an inline comment acknowledging the intentional floating policy.

testutil/ptr.go
```diff
@@ -1,4 +1,6 @@
...
+func Ptr[T any](v T) *T { return new(v) }
```
lugassawan, line 8: **Low — Doc comment predates the `new(expr)` body**

The comment on line 3 says "returns a pointer to v" without mentioning the mechanism. Now that the body uses `new(v)` (a Go 1.26 builtin form), a reader who knows that `new` classically takes a type, not an expression, may do a double-take.

**Suggested fix** (minor): update or extend the comment — e.g. "returns a pointer to v (uses Go 1.26 `new(expr)` form)".

go.mod
lugassawan: Addressed in 6d1c919: ran `go mod tidy` to normalise the module graph under go1.26.3. A bare `toolchain go1.26.3` directive is stripped by `go mod tidy` when it equals the `go` directive (Go spec: identical versions are redundant). The `go 1.26.3` line pins the minimum; CI is further locked via `go-version-file: go.mod` in every `setup-go` step, which overrides `GOTOOLCHAIN=auto` and always installs exactly go1.26.3 in CI regardless of what is available locally.

testutil/ptr.go
lugassawan: Addressed in 6d1c919: extended the doc comment to mention the Go 1.26 `new(expr)` form and the `//go:fix inline` annotation intent, so future maintainers have the context inline.

.github/workflows/ci.yml
lugassawan: Intentional — `@latest` was an explicit decision in the issue plan ("Keep govulncheck `@latest` and `cache: true` — out of scope / unchanged"). The reproducibility gap is narrowed by `go-version-file: go.mod` (pinned toolchain) which is the higher-value lock; govulncheck itself rarely changes its exit-code semantics in patch releases. Happy to pin it in a follow-up if desired.

testutil/ptr.go
lugassawan: Addressed in 6d1c919: doc comment now reads 'Uses Go 1.26 new(expr) form (module minimum ≥ 1.26.3); //go:fix inline lets tooling rewrite call-sites automatically.'

Issue #234, opened 2026-05-31
[feat] Windows self-update: atomic binary replace (rename-aside mechanism)


lugassawan:
## Context

`rimba update` now correctly downloads, verifies (SHA-256 checksum), and extracts the Windows release `.zip` on Windows. However the final self-replace step — swapping the running `.exe` — is not yet supported.

On Windows, a running process holds an exclusive lock on its own `.exe`. The standard Unix `os.Rename` trick fails with **"Access is denied"**. The correct approach is a rename-aside dance:

1. Move the old `.exe` aside to `.exe.old` (Windows allows renaming a running binary)
2. Copy the new binary into place at the original path
3. On failure: restore `.old` (rollback)
4. On next startup: sweep and delete any `.exe.old` left from a previous update

## Current behaviour

`rimba update` on Windows will download and verify the release successfully, then surface a clear `errhint` pointing to this issue instead of attempting the rename (which would crash with a cryptic OS error).

## Scope

- Implement `swapBinary` in `internal/updater/swap_windows.go`
- Expand-phase: rename `rimba.exe` → `rimba.exe.old`, copy new binary → `rimba.exe`, chmod
- Rollback-phase: restore `.old` on any copy/chmod failure
- Startup sweep: at `cmd` init, delete `<executable>.old` if it exists
- Add integration test (Windows CI or cross-compile check)

## Related

- PR that introduced checksum verification and deferred this: #222

PR #235, opened 2026-05-31
[feat] verify self-update integrity with checksum + guard release-asset contract

lugassawan:
## Summary

- Add SHA-256 checksum verification against goreleaser's `checksums.txt` before any binary is swapped — fail-closed on missing file, absent asset row, or hash mismatch
- `DownloadResult` carries `ArchivePath`, `BinaryPath`, and `SHA256` (archive hash, matching what goreleaser signs)
- `FetchChecksums` parses sha256sum(1) format; `verifyChecksumMatch` is a hard error on absent row — no silent skip
- Pipeline in `runner.go` reordered: `check → download → verifyChecksum → prepare → locate → ctx.Err() → install → verify`; `TestRunReplaceNotCalledOnChecksumFailure` encodes the fail-closed invariant
- Windows asset/extract: `Check()` uses `.zip` extension, `Download()` extracts `rimba.exe`; `swap_windows.go` returns a clear `errhint` stub pointing to follow-up #234 rather than crashing with "Access is denied"
- Contract test (`contract_test.go`): reads `.goreleaser.yaml`, renders `name_template` via `text/template` for all 6 `{linux,darwin,windows}×{amd64,arm64}` targets, asserts match with `assetNameFor()` — any goreleaser rename breaks CI immediately
- `checksums.txt` response capped at 1 MiB (defense-in-depth alongside the 100 MiB archive limit)

## Test Plan

- [x] Unit tests pass (`make test`) — 1887/1887
- [x] Linter passes (`make lint`) — 0 issues (gocognit < 15, goconst, perfsprint, tidygo all clean)
- [x] Total coverage ≥ 97% (`go tool cover -func` → 97.0%)

### Type-tailored checks (feat)

- [x] `TestRunReplaceNotCalledOnChecksumFailure` — verifies fail-closed: `replace` seam not called when checksum mismatch
- [x] `TestCheckMissingChecksums` — `Check()` returns error when `checksums.txt` absent from release assets
- [x] `TestVerifyChecksumMatchAbsentRow` — absent asset row is hard error
- [x] `TestVerifyChecksumMatchMismatch` — hash mismatch returns descriptive error
- [x] `TestDownloadZipArchive` — Windows `.zip` correctly extracted as `rimba.exe`
- [x] `TestAssetNameContract` — 6-row table coupling goreleaser template to `assetNameFor()`
- [x] `TestHintSurfaces` includes `verifyChecksum fails` row verifying errhint wiring

## Issue

Closes #222

### DIFF
--- a/internal/updater/checksum.go
+++ b/internal/updater/checksum.go
@@ -1,0 +1,70 @@
+package updater
+
+import (
+	"context"
+	"fmt"
+	"io"
+	"net/http"
+	"strings"
+)
+
+// maxChecksumSize caps the checksums.txt download at 1 MiB — orders of magnitude
+// larger than any real goreleaser checksums file.
+var maxChecksumSize int64 = 1 << 20
+
+// FetchChecksums downloads the checksums file at url and returns a map of
+// filename → lowercase SHA-256 hex. Each line must be in sha256sum(1) format:
+// "<hex>  <filename>" (one or two spaces). Lines that don't parse are skipped.
+func (u *Updater) FetchChecksums(ctx context.Context, url string) (map[string]string, error) {
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, url, nil)
+	if err != nil {
+		return nil, fmt.Errorf("creating checksums request: %w", err)
+	}
+
+	resp, err := u.Client.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("fetching checksums: %w", err)
+	}
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return nil, fmt.Errorf("checksums request returned status %d", resp.StatusCode)
+	}
+
+	body, err := io.ReadAll(io.LimitReader(resp.Body, maxChecksumSize+1))
+	if err != nil {
+		return nil, fmt.Errorf("reading checksums: %w", err)
+	}
+	if int64(len(body)) > maxChecksumSize {
+		return nil, fmt.Errorf("checksums response exceeds %d bytes", maxChecksumSize)
+	}
+
+	sums := make(map[string]string)
+	for line := range strings.SplitSeq(string(body), "\n") {
+		line = strings.TrimSpace(line)
+		if line == "" {
+			continue
+		}
+		fields := strings.Fields(line)
+		if len(fields) < 2 {
+			continue
+		}
+		hex := strings.ToLower(fields[0])
+		filename := fields[len(fields)-1]
+		sums[filename] = hex
+	}
+	return sums, nil
+}
+
+// verifyChecksumMatch checks that gotHex matches the expected SHA-256 for assetName.
+// An absent row is treated as a hard error (fail-closed).
+func verifyChecksumMatch(sums map[string]string, assetName, gotHex string) error {
+	expected, ok := sums[assetName]
+	if !ok {
+		return fmt.Errorf("asset %q not found in checksums.txt", assetName)
+	}
+	if strings.ToLower(gotHex) != expected {
+		return fmt.Errorf("checksum mismatch for %s: got %s, want %s", assetName, gotHex, expected)
+	}
+	return nil
+}
--- a/internal/updater/contract_test.go
+++ b/internal/updater/contract_test.go
@@ -1,0 +1,112 @@
+package updater
+
+import (
+	"bytes"
+	"os"
+	"strings"
+	"testing"
+	"text/template"
+)
+
+// goreleaserField extracts the value of a top-level or nested YAML key by
+// scanning for the first line (after trimming) that starts with "key:".
+// Surrounding quotes are stripped. Returns "" when not found.
+func goreleaserField(content, key string) string {
+	prefix := key + ":"
+	for line := range strings.SplitSeq(content, "\n") {
+		trimmed := strings.TrimSpace(line)
+		if after, ok := strings.CutPrefix(trimmed, prefix); ok {
+			return strings.Trim(strings.TrimSpace(after), `"'`)
+		}
+	}
+	return ""
+}
+
+// archiveNameTemplate returns the first name_template line that contains "{{",
+// distinguishing the archive template from the checksum name_template.
+func archiveNameTemplate(content string) string {
+	prefix := "name_template:"
+	for line := range strings.SplitSeq(content, "\n") {
+		trimmed := strings.TrimSpace(line)
+		if after, ok := strings.CutPrefix(trimmed, prefix); ok && strings.Contains(trimmed, "{{") {
+			return strings.Trim(strings.TrimSpace(after), `"'`)
+		}
+	}
+	return ""
+}
+
+// TestAssetNameContract verifies that the updater's assetNameFor helper produces
+// exactly the filename goreleaser would generate for every release target.
+// Any edit to .goreleaser.yaml's name_template that drifts from updater.go will
+// fail this test in CI before it can silently break rimba update.
+func TestAssetNameContract(t *testing.T) {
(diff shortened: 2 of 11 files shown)


internal/updater/extract.go
```diff
@@ -0,0 +1,106 @@
...
+			retErr = cerr
+		}
+	}()
+
+	lr := io.LimitReader(r, maxBinarySize+1)
+	n, err := io.Copy(f, lr)
+	if err != nil {
+		return "", fmt.Errorf("extracting binary: %w", err)
+	}
+	if n > maxBinarySize {
```
lugassawan, line 101: **Medium** — `os.Remove(dst)` is called at line 101 before the deferred `f.Close()` runs. On Unix the unlinked file's inode stays alive until the FD closes (harmless); on Windows `Remove` on an open file fails, so the oversized binary survives in the temp dir.

Consider closing before removing, or dropping the explicit `os.Remove` and relying on the parent-dir `CleanupTempDir` that already handles the cleanup path:

```go
// Option A: close then remove
n, err := io.Copy(f, lr)
if cerr := f.Close(); err == nil { err = cerr }
...
if n > maxBinarySize {
    _ = os.Remove(dst)
    return "", fmt.Errorf(...)
}
// Option B: drop os.Remove entirely — CleanupTempDir covers it
```

internal/updater/contract_test.go
```diff
@@ -0,0 +1,112 @@
...
+		if after, ok := strings.CutPrefix(trimmed, prefix); ok {
+			return strings.Trim(strings.TrimSpace(after), `"'`)
+		}
+	}
+	return ""
+}
+
+// archiveNameTemplate returns the first name_template line that contains "{{",
+// distinguishing the archive template from the checksum name_template.
+func archiveNameTemplate(content string) string {
```
lugassawan, line 29: **Medium** — `archiveNameTemplate` identifies the archive `name_template` by scanning for the first `name_template:` line containing `{{`. This works because the current YAML has the archive template before the checksum `name_template: checksums.txt`, but it ignores YAML structure. A future change that adds a second template-bearing `name_template:` above the `archives:` block would silently return the wrong template — and the test would still pass while coupling to the wrong value.

One option: scope the search to lines that appear after an `archives:` heading and reset state on other top-level keys. Another: embed the expected template string as a constant and assert equality directly, bypassing the dynamic parse.

internal/updater/checksum.go
```diff
@@ -0,0 +1,70 @@
...
+		return nil, fmt.Errorf("checksums request returned status %d", resp.StatusCode)
+	}
+
+	body, err := io.ReadAll(io.LimitReader(resp.Body, maxChecksumSize+1))
+	if err != nil {
+		return nil, fmt.Errorf("reading checksums: %w", err)
+	}
+	if int64(len(body)) > maxChecksumSize {
+		return nil, fmt.Errorf("checksums response exceeds %d bytes", maxChecksumSize)
+	}
```
lugassawan, line 40: **Low** — `maxChecksumSize` has the same `LimitReader(n+1) + n > limit` guard as `maxArchiveSize` (covered by `TestDownloadExceedsMaxArchiveSize`), but there is no `TestFetchChecksumsExceedsMaxSize` analog. The branch is reachable.

```go
func TestFetchChecksumsExceedsMaxSize(t *testing.T) {
    orig := maxChecksumSize
    maxChecksumSize = 5
    t.Cleanup(func() { maxChecksumSize = orig })

    srv := serveChecksums(t, "aabbccdd  rimba_2.0.0_linux_amd64.tar.gz\n") // >5 bytes
    u := newTestUpdater(srv)

    _, err := u.FetchChecksums(context.Background(), srv.URL+"/checksums.txt")
    if err == nil {
        t.Fatal("expected error when checksums response exceeds size limit")
    }
    if !strings.Contains(err.Error(), "exceeds") {
        t.Errorf("error = %q, want 'exceeds'", err.Error())
    }
}
```

internal/updater/runner_test.go
```diff
@@ -18,6 +18,13 @@ const (
 	testCurrentBinary = "/usr/local/bin/rimba"
 )
 
+// fakeDownloadResult is the DownloadResult returned by the default test download seam.
+var fakeDownloadResult = &DownloadResult{
```
lugassawan: **Low** — `fakeDownloadResult` is a package-level `var`, so a future test that mutates `fakeDownloadResult.SHA256` (e.g., to test a checksum mismatch scenario) would silently bleed into other tests running in the same binary. Currently nothing mutates it, but declaring it as a local inside `newTestRunner` or as a function would make the isolation explicit.

internal/updater/updater.go
```diff
@@ -291,56 +343,25 @@ func EnsurePath(dir string) error {
...
+// findReleaseAssets scans release assets for the platform archive and checksums file.
+// Returns empty strings for any asset not found.
+func findReleaseAssets(assetName string, assets []Asset) (downloadURL, checksumsURL string) {
+	for _, a := range assets {
+		if a.Name == assetName {
+			downloadURL = a.BrowserDownloadURL
+		}
+		if a.Name == checksumsFileName {
+			checksumsURL = a.BrowserDownloadURL
 		}
```
lugassawan, line 367: **Low (informational)** — `BrowserDownloadURL` from the GitHub release JSON is used verbatim as the download and checksums URL with no validation that it points to `github.com` or `objects.githubusercontent.com`. The ticket explicitly calls out "redirected BrowserDownloadURL" as an attack vector — checksum verification is the intended mitigation and it works. But a defence-in-depth URL prefix check would close the TOCTOU window between API fetch and download:

```go
// optional: add to findReleaseAssets or Check()
if !strings.HasPrefix(downloadURL, "https://github.com/") &&
   !strings.HasPrefix(downloadURL, "https://objects.githubusercontent.com/") {
    return nil, fmt.Errorf("unexpected download URL host: %s", downloadURL)
}
```

Alternatively, document this as a conscious decision in a comment.

internal/updater/extract.go
lugassawan: Addressed in de8e1e36fcfcd2a5127885b86f072fddc55a0fff: removed the premature `os.Remove` — closing an open FD on Windows requires the file to be closed first, and `CleanupTempDir` already removes the entire temp dir on the error path.

internal/updater/contract_test.go
lugassawan: Addressed in de8e1e36fcfcd2a5127885b86f072fddc55a0fff: `archiveNameTemplate` now enters the `archives:` section on that top-level key and exits on the next non-indented line — so a future `name_template:` entry elsewhere in the file can't match.

internal/updater/checksum.go
lugassawan: Addressed in de8e1e36fcfcd2a5127885b86f072fddc55a0fff: added `TestFetchChecksumsExceedsMaxSize` which sets `maxChecksumSize = 5`, serves a body longer than 5 bytes, and asserts the 'exceeds' error — symmetric with `TestDownloadExceedsMaxArchiveSize`.

internal/updater/runner_test.go
lugassawan: Addressed in de8e1e36fcfcd2a5127885b86f072fddc55a0fff: inlined the `DownloadResult` literal directly into `newTestRunner`; no longer a package-level var that could bleed between test cases.

internal/updater/updater.go
lugassawan: Addressed in de8e1e36fcfcd2a5127885b86f072fddc55a0fff: added a doc comment to `findReleaseAssets` explaining the conscious decision — checksum verification (issue #222) is the intended mitigation against redirected-URL attacks rather than host-prefix validation.

PR #236, opened 2026-05-31
[feat] thread cancellation context through git, hooks, and dep installs

lugassawan:
## Summary

- **`git.Runner`** gains `RunContext`/`RunInDirContext`; `ExecRunner` uses `exec.CommandContext` as its single primitive so every git subprocess is killable via ctx
- **git wrappers** (`Fetch`, `Push`, `PushForceWithLease`, `Rebase`, `Merge`, `RemotePrune`, `DeleteRemoteBranch`, `PruneRemotes`) accept `ctx` as first param; `AbortRebase` stays non-ctx (rebase recovery must survive Ctrl-C)
- **`deps`**: `RunPostCreateHooks` / `runInstall` switch to `exec.CommandContext`; hook loop breaks early on `ctx.Err()`
- **operations layer**: `SyncBranch`, `SyncWorktree`, `PushBranch`, `AddWorktree`, `PostCreateSetup`, `PostRenameSetup`, `RemoveCandidates`, `MergeWorktree` all carry ctx through
- **cobra commands** (`sync`, `clean`, `add`, `duplicate`, `restore`, `rename`, `deps`, `merge`): thread `cmd.Context()` into all network-touching helpers
- **MCP handlers** (`tool_sync`, `tool_clean`, `tool_add`, `tool_merge`): thread existing handler ctx into all operation calls
- **`debug.TimedRunner`** gains the two ctx methods (time + delegate to inner)
- **12 test fakes** gain delegating `RunContext`/`RunInDirContext` so existing arg-capture assertions keep passing

This is a retrofit of the existing in-repo precedent (`gh.Runner.Run(ctx, args...)`) onto `git.Runner`, not a new design.

## TDD tests added

- `TestRunInDirContextCancelled` — pre-cancelled ctx returns error containing "context canceled"
- `TestRunDelegatesToContext` — non-ctx `Run` still produces output via delegation chain
- `TestFetchPassesContext` — sentinel ctx value flows through to `RunContext`
- `TestRebaseContextCancelledReturnsFast` — fake blocks on `<-ctx.Done()`; cancel @10ms → returns in <1s
- `TestRunPostCreateHooksCancelledStopsLaunching` — pre-cancelled, no hooks launch, marker not created
- `TestRunPostCreateHooksKillsChild` — `sleep 30`, cancel @50ms → returns in <2s with error
- `TestRunInstallCancelledKillsChild` — `InstallCmd="sleep 30"`, cancel mid-flight → fast error

## Deferred follow-ups (noted in #223)

- Configurable per-command timeout (e.g. `--timeout 30s`)
- `parallel.Collect` cancellation propagation
- `fsutil.DirSize` cancellation

## Test plan

- [x] `go test -race ./...` — 1895 passed, 0 failures
- [x] `make lint` — 0 issues (`fatcontext`, `maxparams` ≤7, `thelper` clean)
- [x] `go tool cover -func=coverage.out | grep total` — 96.8% (pre-existing baseline; all new branches covered by TDD above)
- [ ] Manual smoke: `make build && rimba sync --all` + Ctrl-C mid-fetch (git child dies)
- [ ] Worktree with `post_create: ["sleep 30"]` + Ctrl-C (hook subprocess killed)

Closes #223

### DIFF
--- a/cmd/clean.go
+++ b/cmd/clean.go
@@ -3,6 +3,7 @@ package cmd
 import (
 	"bufio"
 	"context"
+	"errors"
 	"fmt"
 	"strings"
 
@@ -223,8 +224,9 @@ func cleanMerged(ctx context.Context, cmd *cobra.Command, r git.Runner) error {
 		summaryFmt:    "Cleaned %d merged worktree(s).\n",
 		originPresent: remotePresent, // shared with printMergedCandidates — single probe
 		preFind: func(c *cobra.Command, rr git.Runner, sp *spinner.Spinner) error {
-			mergeRef = cleanFetchMergeRef(ctx, c, rr, sp, mainBranch)
-			return nil
+			var err error
+			mergeRef, err = cleanFetchMergeRef(ctx, c, rr, sp, mainBranch)
+			return err
 		},
 		find: func(rr git.Runner) ([]operations.CleanCandidate, []string, error) {
 			result, err := operations.FindMergedCandidates(rr, mergeRef, mainBranch)
@@ -272,15 +274,19 @@ func cleanStale(ctx context.Context, cmd *cobra.Command, r git.Runner) error {
 }
 
 // cleanFetchMergeRef fetches from origin and returns the ref to diff against.
-// Falls back to mainBranch with a warning if fetch fails.
-func cleanFetchMergeRef(ctx context.Context, cmd *cobra.Command, r git.Runner, s *spinner.Spinner, mainBranch string) string {
+// Returns an error on cancellation; falls back to mainBranch with a warning on
+// other fetch failures (e.g. no remote configured).
+func cleanFetchMergeRef(ctx context.Context, cmd *cobra.Command, r git.Runner, s *spinner.Spinner, mainBranch string) (string, error) {
 	s.Start("Fetching from " + git.DefaultRemote + "...")
 	if err := git.Fetch(ctx, r, git.DefaultRemote); err != nil {
 		s.Stop()
+		if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
+			return "", err
+		}
 		fmt.Fprintf(cmd.ErrOrStderr(), "Warning: fetch failed (no remote?): continuing with local state\n")
-		return mainBranch
+		return mainBranch, nil
 	}
-	return git.DefaultRemote + "/" + mainBranch
+	return git.DefaultRemote + "/" + mainBranch, nil
 }
 
 func cleanRemoveCandidates(ctx context.Context, cmd *cobra.Command, r git.Runner, s *spinner.Spinner, candidates []operations.CleanCandidate, originPresent bool) int {
--- a/cmd/clean.go
+++ b/cmd/clean.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"bufio"
+	"context"
 	"fmt"
 	"strings"
 
@@ -39,11 +40,11 @@ var cleanCmd = &cobra.Command{
 
 		switch {
 		case merged:
-			return cleanMerged(cmd, r)
+			return cleanMerged(cmd.Context(), cmd, r)
 		case stale:
-			return cleanStale(cmd, r)
+			return cleanStale(cmd.Context(), cmd, r)
 		default:
-			return cleanPrune(cmd, r)
+			return cleanPrune(cmd.Context(), cmd, r)
 		}
 	},
 }
@@ -60,7 +61,7 @@ func init() {
 	rootCmd.AddCommand(cleanCmd)
 }
 
-func cleanPrune(cmd *cobra.Command, r git.Runner) error {
+func cleanPrune(ctx context.Context, cmd *cobra.Command, r git.Runner) error {
 	dryRun, _ := cmd.Flags().GetBool(flagDryRun)
 
 	hint.New(cmd, hintPainter(cmd)).
@@ -87,12 +88,12 @@ func cleanPrune(cmd *cobra.Command, r git.Runner) error {
 		fmt.Fprintln(cmd.OutOrStdout(), "Pruned stale worktree references.")
 	}
 
-	return cleanRemotePrune(cmd, r, s, dryRun)
+	return cleanRemotePrune(ctx, cmd, r, s, dryRun)
 }
 
 // cleanRemotePrune prunes stale remote-tracking refs across all configured remotes.
 // Skips gracefully when there are no remotes; warns and continues on per-remote failure.
-func cleanRemotePrune(cmd *cobra.Command, r git.Runner, s *spinner.Spinner, dryRun bool) error {
+func cleanRemotePrune(ctx context.Context, cmd *cobra.Command, r git.Runner, s *spinner.Spinner, dryRun bool) error {
 	s.Start("Pruning remote-tracking refs...")
 	remotes, err := git.ListRemotes(r)
 	if err != nil {
@@ -104,7 +105,7 @@ func cleanRemotePrune(cmd *cobra.Command, r git.Runner, s *spinner.Spinner, dryR
 		fmt.Fprintln(cmd.OutOrStdout(), "No remotes; skipped remote-ref prune.")
 		return nil
 	}
-	pruned, failures := git.PruneRemotes(r, remotes, dryRun)
+	pruned, failures := git.PruneRemotes(ctx, r, remotes, dryRun)
 	s.Stop()
 	failureMsgs := make([]string, len(failures))
 	for i, f := range failures {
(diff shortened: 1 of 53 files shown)


internal/deps/hooks.go
```diff
@@ -16,13 +17,16 @@ type HookResult struct {
 }
 
 // RunPostCreateHooks executes shell commands in the worktree directory.
-// It collects errors but does not stop on failure — all hooks run regardless.
-func RunPostCreateHooks(worktreeDir string, hooks []string, onProgress progress.Func) []HookResult {
+// It stops launching new hooks if ctx is cancelled, but does not stop an already-running hook.
```
lugassawan: **High** — Doc comment contradicts runtime behaviour

The comment says "does not stop an already-running hook" but `exec.CommandContext` sends SIGKILL when ctx is cancelled — `TestRunPostCreateHooksKillsChild` confirms this.

```go
// RunPostCreateHooks executes shell commands in the worktree directory.
// Stops launching new hooks if ctx is cancelled; kills any already-running
// hook when ctx is cancelled (via exec.CommandContext).
```

cmd/sync.go
```diff
@@ -91,7 +92,7 @@ var syncCmd = &cobra.Command{
 			fmt.Fprintln(cmd.OutOrStdout(), "[dry-run] would fetch origin")
 		} else {
 			s.Start("Fetching from origin...")
-			if err := git.Fetch(r, "origin"); err != nil {
+			if err := git.Fetch(cmd.Context(), r, "origin"); err != nil {
 				s.Stop()
 				fmt.Fprintf(cmd.ErrOrStderr(), "Warning: fetch failed (no remote?): continuing with local state\n")
```
lugassawan, line 100: **Medium** — `context.Canceled` from Ctrl-C is silently treated as "no remote"

When the user hits Ctrl-C mid-fetch, `git.Fetch` returns `context.Canceled`. All non-fatal fetch call sites swallow every error identically, print the misleading "fetch failed (no remote?)" warning, and continue with local state. The same pattern repeats in `cmd/clean.go:278`, `internal/mcp/tool_sync.go:70`, and `internal/mcp/tool_clean.go:93`.

```go
if err := git.Fetch(cmd.Context(), r, "origin"); err != nil {
    s.Stop()
    if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
        return err
    }
    fmt.Fprintf(cmd.ErrOrStderr(), "Warning: fetch failed (no remote?): continuing with local state\n")
}
```

cmd/clean.go
```diff
@@ -272,19 +273,19 @@ func cleanStale(cmd *cobra.Command, r git.Runner) error {
 
 // cleanFetchMergeRef fetches from origin and returns the ref to diff against.
 // Falls back to mainBranch with a warning if fetch fails.
-func cleanFetchMergeRef(cmd *cobra.Command, r git.Runner, s *spinner.Spinner, mainBranch string) string {
+func cleanFetchMergeRef(ctx context.Context, cmd *cobra.Command, r git.Runner, s *spinner.Spinner, mainBranch string) string {
 	s.Start("Fetching from " + git.DefaultRemote + "...")
-	if err := git.Fetch(r, git.DefaultRemote); err != nil {
+	if err := git.Fetch(ctx, r, git.DefaultRemote); err != nil {
 		s.Stop()
 		fmt.Fprintf(cmd.ErrOrStderr(), "Warning: fetch failed (no remote?): continuing with local state\n")
```
lugassawan, line 286: **Medium** — Same fetch-cancellation masking as `cmd/sync.go:97`

`cleanFetchMergeRef` swallows `context.Canceled` (Ctrl-C) and silently falls back to the local `mainBranch` ref, so `rimba clean --merged` after Ctrl-C will evaluate merges against the local ref rather than the remote ref and may silently leave merged worktrees behind.

cmd/sync.go
```diff
@@ -184,7 +185,7 @@ func syncOne(sc *syncContext, input string, worktrees []resolver.WorktreeInfo, p
 	return nil
 }
 
-func syncAll(sc *syncContext, worktrees []resolver.WorktreeInfo, prefixes []string, useMerge, includeInherited, push bool) error { //nolint:unparam // error return matches RunE contract
+func syncAll(ctx context.Context, sc *syncContext, worktrees []resolver.WorktreeInfo, prefixes []string, useMerge, includeInherited, push bool) error { //nolint:unparam // error return matches RunE contract
```
lugassawan, line 191: **Medium** — `syncAll` prints a partial success summary after Ctrl-C

Now that `syncAll` receives `ctx`, it should check `ctx.Err()` after `wg.Wait()` to avoid printing a misleading summary (e.g. "Rebased 3 worktrees onto main") for the subset that completed before the user cancelled.

```go
wg.Wait()
sc.s.Stop()
if ctx.Err() != nil {
    return ctx.Err()   // skip summary; cancellation was intentional
}
if !sc.dryRun {
    printSyncSummary(sc.cmd, sc.cfg.DefaultSource, useMerge, sc.res)
}
```

internal/deps/hooks.go
```diff
@@ -16,13 +17,16 @@ type HookResult struct {
...
+func RunPostCreateHooks(ctx context.Context, worktreeDir string, hooks []string, onProgress progress.Func) []HookResult {
 	results := make([]HookResult, 0, len(hooks))
 	for i, hook := range hooks {
+		if ctx.Err() != nil {
```
lugassawan, line 26: **Medium** — `ctx.Err()` pre-check comment overstates its role

The `if ctx.Err() != nil { break }` check (a micro-optimisation that avoids *launching* a new hook) is silently in tension with the doc comment corrected in the adjacent finding: the real cancellation mechanism is `exec.CommandContext`, not this guard. Once the doc comment is updated to say `exec.CommandContext` kills the child, this guard can simply be commented as an early-exit optimisation, not the primary cancellation path.

internal/git/sync_ctx_test.go
```diff
@@ -0,0 +1,92 @@
...
+	m.capturedCtx = ctx
+	if m.runInDir != nil {
+		return m.runInDir(ctx, dir, args...)
+	}
+	return "", nil
+}
+
+type ctxKey struct{}
+
+func TestFetchPassesContext(t *testing.T) {
```
lugassawan, line 61: **Low** — `TestFetchPassesContext` verifies ctx plumbing through a mock, not real subprocess kill

The test confirms the sentinel ctx value reaches `RunContext` via `ctxMockRunner`, but it does not verify that a real `git fetch` subprocess is actually killed. Compare with `TestRunInDirContextCancelled` in `git_ctx_test.go`, which uses `ExecRunner` against a real git binary. Consider a follow-up test that cancels a real `git fetch` to complement the kill coverage already present for hooks and runInstall.

internal/deps/hooks.go
lugassawan: test

lugassawan: Addressed in 3428a1441f3047283678aaac91d4900bd08ccca3: corrected doc comment — exec.CommandContext kills in-flight hooks on cancellation; the ctx.Err() pre-check is an early-launch optimisation, not the primary kill mechanism.

cmd/sync.go
lugassawan: Addressed in 3428a1441f3047283678aaac91d4900bd08ccca3: added errors.Is(err, context.Canceled/DeadlineExceeded) check in cmd/sync.go; cancellation now propagates instead of emitting the misleading 'no remote?' warning.

cmd/clean.go
lugassawan: Addressed in 3428a1441f3047283678aaac91d4900bd08ccca3: cleanFetchMergeRef now returns (string, error); cancellation propagates via preFind so 'rimba clean --merged' exits cleanly after Ctrl-C instead of silently diffing against local state.

cmd/sync.go
lugassawan: Addressed in 3428a1441f3047283678aaac91d4900bd08ccca3: added ctx.Err() check after wg.Wait() in syncAll; a cancelled run now returns the error and skips the partial-success summary.

internal/deps/hooks.go
lugassawan: Addressed in 3428a1441f3047283678aaac91d4900bd08ccca3: doc comment on RunPostCreateHooks updated to accurately describe both the early-launch guard (ctx.Err() check) and the subprocess-kill behaviour (exec.CommandContext).

internal/git/sync_ctx_test.go
lugassawan: Addressed in 9f99a6377ca83e00c25dcfa6161b87aff93e155a: added TestFetchContextCancelledReturnsFast which uses ExecRunner directly (not a mock) and verifies a pre-cancelled ctx causes Fetch to return context.Canceled immediately. Also added TestRunContextDelegatesToRunInDirContext to cover the ExecRunner.RunContext delegation path.

PR #238, opened 2026-05-31
[fix] harden concurrency: Collect auto-guard, --concurrency validation, filterDirtyWorktrees refactor

lugassawan:
## Summary

- **`parallel.Collect` self-deadlock guard**: `make(chan struct{}, 0)` is unbuffered; when `concurrency=0` and `n>0` the first goroutine blocks on send forever. Added `if concurrency <= 0 { concurrency = n }` after the `n==0` early return, mirroring `executor.go:49`.
- **`--concurrency` flag validation**: `rimba exec --concurrency -5` was silently coerced to "unlimited" by the executor's `<=0` guard. Now rejects `< 0` in `execValidateFlags` with an `errhint.WithFix` actionable error; `0` stays valid ("unlimited").
- **`filterDirtyWorktrees` refactor**: replaced hand-rolled `sync.WaitGroup + chan struct{}` semaphore with `parallel.Collect[dirtyResult]`, preserving the same behavior (error-as-dirty, warning to stderr, order-preserving output, fixed bound of 8).

## Test Plan

- [x] Race detector clean (`make test-race`)
- [x] Coverage ≥ 97% (`make test-coverage` → 97.1%)

## Issue

Closes #224

### DIFF
--- a/cmd/exec.go
+++ b/cmd/exec.go
@@ -5,7 +5,7 @@ import (
 	"fmt"
 	"io"
 	"strings"
-	"sync"
+	"sync/atomic"
 
 	"github.com/lugassawan/rimba/internal/config"
 	"github.com/lugassawan/rimba/internal/errhint"
@@ -14,6 +14,7 @@ import (
 	"github.com/lugassawan/rimba/internal/hint"
 	"github.com/lugassawan/rimba/internal/operations"
 	"github.com/lugassawan/rimba/internal/output"
+	"github.com/lugassawan/rimba/internal/parallel"
 	"github.com/lugassawan/rimba/internal/resolver"
 	"github.com/lugassawan/rimba/internal/spinner"
 	"github.com/lugassawan/rimba/internal/termcolor"
@@ -117,6 +118,12 @@ func execValidateFlags(opts execOpts) error {
 	if err := validateTypeFilter(opts.typeFilter); err != nil {
 		return err
 	}
+	if opts.concurrency < 0 {
+		return errhint.WithFix(
+			errors.New("--concurrency must be >= 0"),
+			"run: rimba exec --concurrency <n>  (n >= 0; 0 = unlimited)",
+		)
+	}
 	return nil
 }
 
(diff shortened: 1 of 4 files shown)


cmd/exec.go
```diff
@@ -218,46 +225,40 @@ func init() {
...
-			}
-		}(i, wt.Path)
-	}
-	wg.Wait()
+	n := len(worktrees)
+	var done atomic.Int32
+
+	results := parallel.Collect[dirtyResult](n, 8, func(i int) dirtyResult {
+		count := done.Add(1)
+		s.Update(fmt.Sprintf("Checking dirty status... (%d/%d)", count, n))
```
lugassawan: **Medium** — Spinner progress counter is non-monotonic under concurrency.

`done.Add(1)` is called at goroutine-start, not goroutine-completion, so when the concurrency cap (8) is hit all 8 goroutines atomically increment `done` before any `IsDirty` call completes. The display can jump to `8/n` in a single render tick and show out-of-order values (e.g. `3/5` then `1/5`). The original code incremented sequentially in the dispatch loop, giving smooth monotonic feedback.

Acknowledged as a known trade-off in the PR description. Consider moving the increment to after the `git.IsDirty` call returns to count completions rather than starts — same `atomic.Int32`, still accurate, monotonically increasing.

cmd/exec_test.go
```diff
@@ -353,6 +353,24 @@ func TestExecValidateFlagsOK(t *testing.T) {
...
+
+func TestExecValidateFlagsNegativeConcurrency(t *testing.T) {
+	err := execValidateFlags(execOpts{all: true, concurrency: -1})
+	if err == nil {
+		t.Fatal("expected error for negative concurrency")
+	}
+	if !strings.Contains(err.Error(), "concurrency") {
+		t.Errorf("error %q does not mention 'concurrency'", err.Error())
+	}
 }
```
lugassawan, line 377: **Low** — `TestExecValidateFlagsNegativeConcurrency` asserts that the error message contains `"concurrency"` but does not verify the actionable fix hint is present.

Other validation tests in the suite verify the `errhint` fix string. Without this check, a regression from `errhint.WithFix` back to `errors.New` would pass the test.

Consider adding: `if hint := errhint.HintFrom(err); hint == "" { t.Error("expected fix hint") }` (or whatever accessor the project uses for errhint).

internal/parallel/parallel_test.go
```diff
@@ -33,6 +33,25 @@ func TestCollectZeroItems(t *testing.T) {
...
+			t.Errorf("results[%d] = %d, want %d", i, v, i)
+		}
+	}
+
+	// negative concurrency also treated as auto.
+	results2 := parallel.Collect(3, -1, func(i int) int { return i * 10 })
+	if len(results2) != 3 {
+		t.Fatalf("got %d results, want 3", len(results2))
+	}
+}
```
lugassawan, line 58: **Low** — `TestCollectAutoConcurrency`'s negative-concurrency sub-case only checks `len(results2) == 3` but not the values.

The positive-concurrency sub-case (lines 38–46) checks both length and per-element values. The negative-concurrency path could silently return a zero-valued slice of the right length while the function is broken.

Add a value-check loop to `results2` mirroring the `results` loop above (e.g. `if results2[i] != i*10 { ... }`).

cmd/exec.go
lugassawan: Addressed in 2432b2a: moved `done.Add(1)` to after `git.IsDirty` returns so the spinner counter reflects completions (monotonically increasing) rather than goroutine starts.

cmd/exec_test.go
lugassawan: Addressed in 2432b2a: added `strings.Contains(err.Error(), "To fix:")` assertion to verify the `errhint.WithFix` hint is present.

internal/parallel/parallel_test.go
lugassawan: Addressed in 2432b2a: added per-element value check loop for `results2` mirroring the `results` loop above.

PR #240, opened 2026-06-01
[fix] Fix silent dependency-dir corruption in the CoW clone fallback

lugassawan:
## Summary

- `cowCopy` previously ran `cp -R src dst` as a fallback without first removing any partial `dst` left by a failed CoW attempt; POSIX `cp` nests `src` inside an existing `dst/<basename(src)>`, producing a silently corrupt `node_modules`/`vendor`/`target` tree
- Neither the CoW command nor the fallback captured stderr, so failures surfaced as bare `exit status 1` with the original error dropped
- Fix: introduce `cowCopyCmd` as a package-level var (test seam), call `CombinedOutput()` on both attempts, `os.RemoveAll(dst)` before the fallback, and surface both error causes via `errors.Join` when both fail; a `cmdErr` helper avoids the double-colon when stderr is empty

## Test Plan

Three new regression tests:
- `TestCowCopyFallbackNotNested` — proves non-nested layout after debris cleanup
- `TestCowCopyDoubleFailurePreservesBothCauses` — proves both error causes are preserved
- `TestCowCopyRemoveAllError` — proves RemoveAll failure is surfaced (covers clean-dst branch for ≥97% threshold)

## Issue

Closes #226

### DIFF
--- a/internal/deps/clone.go
+++ b/internal/deps/clone.go
@@ -11,13 +11,18 @@ import (
 	"strings"
 )
 
+const (
+	goosDarwin = "darwin"
+	goosLinux  = "linux"
+)
+
 // cowCopyCmd builds the copy-on-write copy command for the host OS.
 // A package var so tests can inject a failing first-copy seam.
 var cowCopyCmd = func(src, dst string) *exec.Cmd {
 	switch runtime.GOOS {
-	case "darwin":
+	case goosDarwin:
 		return exec.Command("cp", "-c", "-R", src, dst)
-	case "linux":
+	case goosLinux:
 		return exec.Command("cp", "--reflink=auto", "-R", src, dst)
 	default:
 		return exec.Command("cp", "-R", src, dst)
@@ -150,7 +155,7 @@ func cowCopy(src, dst string) error {
 	// Only attempt the fallback on platforms where CoW was actually tried.
 	// On the default branch cowCopyCmd already emits a plain cp -R, so
 	// retrying an identical command on failure adds no value.
-	if runtime.GOOS != "darwin" && runtime.GOOS != "linux" {
+	if runtime.GOOS != goosDarwin && runtime.GOOS != goosLinux {
 		return cowErr
 	}
 
--- a/internal/deps/clone.go
+++ b/internal/deps/clone.go
@@ -8,7 +8,21 @@ import (
 	"os/exec"
 	"path/filepath"
 	"runtime"
+	"strings"
 )
+
+// cowCopyCmd builds the copy-on-write copy command for the host OS.
+// A package var so tests can inject a failing first-copy seam.
+var cowCopyCmd = func(src, dst string) *exec.Cmd {
+	switch runtime.GOOS {
+	case "darwin":
+		return exec.Command("cp", "-c", "-R", src, dst)
+	case "linux":
+		return exec.Command("cp", "--reflink=auto", "-R", src, dst)
+	default:
+		return exec.Command("cp", "-R", src, dst)
+	}
+}
 
 // CloneDir copies a directory from src to dst using CoW (copy-on-write) when available.
 // Falls back to regular copy if CoW is not supported.
@@ -119,23 +133,28 @@ func cloneExtraDirs(srcWT, dstWT string, extraDirs []string) error {
 	return nil
 }
 
+func cmdErr(prefix string, out []byte, err error) error {
+	if msg := strings.TrimSpace(string(out)); msg != "" {
+		return fmt.Errorf("%s: %s: %w", prefix, msg, err)
+	}
+	return fmt.Errorf("%s: %w", prefix, err)
+}
+
 func cowCopy(src, dst string) error {
-	var cmd *exec.Cmd
-	switch runtime.GOOS {
-	case "darwin":
-		cmd = exec.Command("cp", "-c", "-R", src, dst)
-	case "linux":
-		cmd = exec.Command("cp", "--reflink=auto", "-R", src, dst)
-	default:
-		cmd = exec.Command("cp", "-R", src, dst)
+	out, err := cowCopyCmd(src, dst).CombinedOutput()
+	if err == nil {
+		return nil
+	}
+	cowErr := cmdErr("cow copy", out, err)
+
+	// A failed CoW cp can leave a partially written dst; remove it so the
+	// fallback lands src's contents AS dst rather than nested inside dst/<base>.
+	if rmErr := os.RemoveAll(dst); rmErr != nil {
+		return errors.Join(cowErr, fmt.Errorf("clean dst before fallback: %w", rmErr))
 	}
 
-	if err := cmd.Run(); err != nil {
-		if runtime.GOOS == "darwin" || runtime.GOOS == "linux" {
-			fallback := exec.Command("cp", "-R", src, dst)
-			return fallback.Run()
-		}
-		return err
+	if out, fbErr := exec.Command("cp", "-R", src, dst).CombinedOutput(); fbErr != nil {
+		return errors.Join(cowErr, cmdErr("fallback copy", out, fbErr))
 	}
 	return nil
 }
--- a/internal/deps/clone.go
+++ b/internal/deps/clone.go
@@ -147,6 +147,13 @@ func cowCopy(src, dst string) error {
 	}
 	cowErr := cmdErr("cow copy", out, err)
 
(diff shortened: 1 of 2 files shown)


internal/deps/clone.go
```diff
@@ -119,23 +133,28 @@ func cloneExtraDirs(srcWT, dstWT string, extraDirs []string) error {
...
-			return fallback.Run()
-		}
-		return err
+	// A failed CoW cp can leave a partially written dst; remove it so the
+	// fallback lands src's contents AS dst rather than nested inside dst/<base>.
+	if rmErr := os.RemoveAll(dst); rmErr != nil {
+		return errors.Join(cowErr, fmt.Errorf("clean dst before fallback: %w", rmErr))
+	}
+
+	if out, fbErr := exec.Command("cp", "-R", src, dst).CombinedOutput(); fbErr != nil {
```
lugassawan, line 168: **Medium:** On the `default` OS branch, `cowCopyCmd` already emits a plain `cp -R` (no CoW flags). When that first copy fails, `cowCopy` unconditionally removes `dst` and runs a second identical `cp -R` as a "fallback" — doubling the I/O and latency for no gain, since the same command will likely fail for the same reason (permissions, disk full).

The original code avoided this by gating the fallback on `runtime.GOOS == "darwin" || runtime.GOOS == "linux"`. Consider restoring that gate:

```go
// Only fall back on platforms where CoW was actually attempted.
if runtime.GOOS != "darwin" && runtime.GOOS != "linux" {
    return cowErr
}
```

This is a design question, not a correctness bug — the fallback is safe but wasteful on unsupported platforms.

internal/deps/clone_test.go
```diff
@@ -570,6 +571,88 @@ func TestCloneModuleRecursiveExtraDirError(t *testing.T) {
...
+// TestCowCopyRemoveAllError covers the branch where dst cleanup fails after a CoW failure.
+func TestCowCopyRemoveAllError(t *testing.T) {
+	parent := t.TempDir()
+	dst := filepath.Join(parent, "target")
+	if err := os.MkdirAll(dst, 0755); err != nil {
+		t.Fatal(err)
+	}
+
+	// Make parent read-only so os.RemoveAll(dst) fails.
+	if err := os.Chmod(parent, 0555); err != nil {
```
lugassawan, line 659: **Low:** `os.Chmod(parent, 0555)` does not prevent deletion when the process runs as root (uid=0). In containerised CI where tests run as root, `os.RemoveAll(dst)` will succeed and the test will emit a false failure on `"expected error when RemoveAll cannot delete dst"`. The same pattern exists in `TestCloneDirRemoveAllError` (line 474), so this is pre-existing — just repeated here.

Consider adding a root guard for consistency:

```go
if os.Getuid() == 0 {
    t.Skip("chmod-based removal-block does not apply under root")
}
```

internal/deps/clone_test.go
```diff
@@ -570,6 +571,88 @@ func TestCloneModuleRecursiveExtraDirError(t *testing.T) {
...
+// and the fallback cp fail, the returned error names both causes.
+func TestCowCopyDoubleFailurePreservesBothCauses(t *testing.T) {
+	dst := filepath.Join(t.TempDir(), "dst")
+
+	orig := cowCopyCmd
+	cowCopyCmd = func(s, d string) *exec.Cmd { return exec.Command("false") }
+	t.Cleanup(func() { cowCopyCmd = orig })
+
+	// Nonexistent src causes the fallback cp -R to also fail.
+	err := cowCopy("/nonexistent/does/not/exist/src", dst)
```
lugassawan, line 632: **Low:** The test relies on `cp -R /nonexistent/does/not/exist/src dst` failing to trigger the fallback-failure branch. The intent is correct, but the mechanism is implicit — a quirky overlay filesystem that materialises the path would allow the fallback to succeed and the test would pass without exercising the double-failure path. This is unlikely in practice; noting it as a documentation gap. Consider a brief comment:

```go
// /nonexistent/... is guaranteed absent; cp -R will fail, exercising the fallback-error branch.
```

internal/deps/clone.go
lugassawan: Addressed in 6f1a86a2c745693c03c06a34145d57ee1e5aaa42: restored the `darwin/linux` OS gate before `RemoveAll` + fallback so the default-branch path exits immediately on first failure instead of running an identical `cp -R` twice.

internal/deps/clone_test.go
lugassawan: Addressed in 6f1a86a2c745693c03c06a34145d57ee1e5aaa42: added `os.Getuid() == 0` t.Skip guard to both `TestCowCopyRemoveAllError` and the pre-existing `TestCloneDirRemoveAllError` (line 461) for consistency.

internal/deps/clone_test.go
lugassawan: Addressed in 6f1a86a2c745693c03c06a34145d57ee1e5aaa42: replaced the implicit comment with an explicit explanation that `/nonexistent/...` is guaranteed absent on any POSIX host, making the fallback-failure intent clear.

internal/deps/clone_test.go
```diff
@@ -570,6 +574,92 @@ func TestCloneModuleRecursiveExtraDirError(t *testing.T) {
 	}
 }
 
+// TestCowCopyFallbackNotNested verifies that a partial CoW failure (debris left at dst)
+// is corrected before the fallback: dst is re-cleaned so src lands AS dst, not nested.
+func TestCowCopyFallbackNotNested(t *testing.T) {
```
lugassawan, line 580: **Medium (follow-up):** The OS gate added at `clone.go:153` (`runtime.GOOS != "darwin" && runtime.GOOS != "linux"`) returns early with `cowErr` before reaching the `RemoveAll`+fallback block. On any non-darwin/linux host (Windows CI, plan9, etc.), both this test and `TestCowCopyDoubleFailurePreservesBothCauses` (line 616) break:

- `TestCowCopyFallbackNotNested` asserts `cowCopy` returns `nil` — but the OS gate means `cowCopy` returns the CoW error, never attempting the fallback. The test fails with an unexpected error.
- `TestCowCopyDoubleFailurePreservesBothCauses` asserts `"fallback copy"` appears in the error — on non-darwin/linux only `"cow copy"` is present. The test fails.

Add an OS guard at the top of each of these two tests (and `TestCowCopyRemoveAllError`) to match the production gate:

```go
if runtime.GOOS != "darwin" && runtime.GOOS != "linux" {
    t.Skip("CoW fallback only attempted on darwin/linux")
}
```

internal/deps/clone_test.go
```diff
@@ -570,6 +574,92 @@ func TestCloneModuleRecursiveExtraDirError(t *testing.T) {
...
+	if !strings.Contains(err.Error(), "cow copy") {
+		t.Errorf("expected 'cow copy' in error, got: %v", err)
+	}
+	if !strings.Contains(err.Error(), "fallback copy") {
+		t.Errorf("expected 'fallback copy' in error, got: %v", err)
+	}
+}
+
+// TestCowCopyRemoveAllError covers the branch where dst cleanup fails after a CoW failure.
+func TestCowCopyRemoveAllError(t *testing.T) {
```
lugassawan, line 645: **Low (follow-up):** On non-darwin/linux, `cowCopy` returns early at the OS gate before `os.RemoveAll` is ever called, so the chmod-0555 setup in this test is never exercised. The test asserts `err != nil` — which is satisfied by the `cowErr` returned from the OS-gate exit — giving a false green. The test appears to pass, but the RemoveAll branch is not covered.

Apply the same OS guard as the other two `TestCowCopy*` tests so this only runs where the branch is actually reachable.

internal/deps/clone_test.go
lugassawan: Addressed in 5daa626bc2a36e12e510840ea3b9660bce8e0b07: added `runtime.GOOS != goosDarwin && runtime.GOOS != goosLinux` t.Skip guard to all three `TestCowCopy*` tests. Also extracted `goosDarwin`/`goosLinux` constants (required by the `goconst` linter — 5 occurrences each across production and test code).

internal/deps/clone_test.go
lugassawan: Addressed in 5daa626bc2a36e12e510840ea3b9660bce8e0b07: same OS guard applied to `TestCowCopyRemoveAllError` — the test now skips on non-darwin/linux where the RemoveAll branch is unreachable.

PR #244, opened 2026-06-01
[feat] add Gradle CoW clone preset

lugassawan:
## Summary

- Add four `CloneOnly` preset rows for Gradle lockfiles (`settings.gradle`, `settings.gradle.kts`, `build.gradle`, `build.gradle.kts`) — clones project-local `.gradle/` and `build/` from a sibling worktree when lockfile content hashes match
- Add `DirGradle`, `DirGradleBuildOutput`, and four `LockfileGradle*` constants to `internal/deps/module.go`; `settings.*` ordered before `build.*` so multi-project roots are preferred via the existing `seenDirs` first-match-wins mechanism
- Backfill stale Cargo, uv, and poetry rows in `docs/configuration.md` (added in #239 but undocumented); add design trade-off note explaining that global `~/.gradle/caches` and Maven `~/.m2` are out of scope and Maven `target/` support is deferred

**Design trade-off:** rimba's CoW model clones dep dirs from a sibling worktree — no env-var injection or symlink infrastructure exists. A stale `.gradle/` clone is a harmless warm cache: Gradle re-validates via content hashes on next invocation. Global caches (`~/.gradle/caches`) and Maven's `~/.m2` require net-new subsystems and are explicitly deferred.

## Test Plan

- [x] E2E tests pass (`make test-e2e`) — `TestAddWithDepsCloneGradle` covers `build.gradle`, `build.gradle.kts`, and `settings.gradle` end-to-end
- [x] Coverage ≥ 97% (`make test-coverage` — 97.0%)

## Issue

Closes #241

### DIFF
--- a/internal/deps/module_test.go
+++ b/internal/deps/module_test.go
@@ -429,6 +429,91 @@ func TestMatchPresetsInSubdirSeenDirSkip(t *testing.T) {
 	}
 }
 
+func TestDetectModulesGradleEachTrigger(t *testing.T) {
+	triggers := []struct {
+		lockfile string
+	}{
+		{LockfileGradleSettings},
+		{LockfileGradleSettingsKts},
+		{LockfileGradle},
+		{LockfileGradleKts},
+	}
+
+	for _, tc := range triggers {
+		t.Run(tc.lockfile, func(t *testing.T) {
+			dir := t.TempDir()
+			writeFile(t, dir, tc.lockfile, "# gradle")
+
+			modules, err := DetectModules(dir, "")
+			if err != nil {
+				t.Fatal(err)
+			}
+
+			assertModuleCount(t, modules, 1)
+			m := modules[0]
+
+			if m.Dir != DirGradle {
+				t.Errorf(fmtExpectedGot, DirGradle, m.Dir)
+			}
+			if len(m.ExtraDirs) != 1 || m.ExtraDirs[0] != DirGradleBuildOutput {
+				t.Errorf("expected ExtraDirs=[%s], got %v", DirGradleBuildOutput, m.ExtraDirs)
+			}
+			if !m.CloneOnly {
+				t.Error("expected CloneOnly=true for Gradle")
+			}
+			if m.Recursive {
+				t.Error("expected Recursive=false for Gradle")
+			}
+			if m.InstallCmd != "" {
+				t.Errorf("expected empty InstallCmd, got %s", m.InstallCmd)
+			}
+		})
+	}
+}
+
+func TestDetectModulesGradleSettingsWinsOverBuild(t *testing.T) {
+	dir := t.TempDir()
+	writeFile(t, dir, LockfileGradleSettings, "# settings")
+	writeFile(t, dir, LockfileGradle, "# build")
+
+	modules, err := DetectModules(dir, "")
+	if err != nil {
+		t.Fatal(err)
+	}
+
+	assertModuleCount(t, modules, 1)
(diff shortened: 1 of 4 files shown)


tests/e2e/deps_test.go
```diff
@@ -459,3 +459,54 @@ func assertVenvCloneRewritesPaths(t *testing.T, lockfile, task1, task2 string) {
...
+
+	cases := []struct {
+		lockfile string
+		task1    string
+		task2    string
+	}{
+		{deps.LockfileGradle, "gradle-1", "gradle-2"},
+		{deps.LockfileGradleKts, "gradle-kts-1", "gradle-kts-2"},
+		{deps.LockfileGradleSettings, "gradle-settings-1", "gradle-settings-2"},
+	}
```
lugassawan, line 477: **Medium** — E2E test table omits `LockfileGradleSettingsKts`.

The four unit tests in `TestDetectModulesGradleEachTrigger` cover all four Gradle lockfile triggers; the E2E table covers only three. `settings.gradle.kts` is the most common Kotlin DSL entrypoint in modern multi-project Gradle builds and is the only trigger without an integration-level smoke test. A regression in that detection path would go undetected.

```go
{deps.LockfileGradleSettingsKts, "gradle-settings-kts-1", "gradle-settings-kts-2"},
```

internal/deps/module_test.go
```diff
@@ -429,6 +429,91 @@ func TestMatchPresetsInSubdirSeenDirSkip(t *testing.T) {
...
+
+	if m.Dir != filepath.Join(subproject, DirGradle) {
+		t.Errorf("expected %s, got %s", filepath.Join(subproject, DirGradle), m.Dir)
+	}
+	if m.WorkDir != subproject {
+		t.Errorf("expected WorkDir=%s, got %s", subproject, m.WorkDir)
+	}
+	if !m.CloneOnly {
+		t.Error("expected CloneOnly=true for nested Gradle")
+	}
```
lugassawan, line 514: **Low** — `TestDetectModulesGradleNestedSubdir` does not assert that `ExtraDirs` are prefixed correctly for subdirectory modules.

`moduleFromPreset` calls `prefixDirs(subdir, p.ExtraDirs)`, which should produce `["subproject/build"]`. If that call were accidentally skipped for Gradle's `ExtraDirs`, the clone path would copy from the worktree root's `build/` instead of `subproject/build/` — silently copying wrong content or missing it.

Extend the test with:
```go
wantExtra := filepath.Join(subproject, DirGradleBuildOutput)
if len(m.ExtraDirs) != 1 || m.ExtraDirs[0] != wantExtra {
    t.Errorf("expected ExtraDirs=[%s], got %v", wantExtra, m.ExtraDirs)
}
```

tests/e2e/deps_test.go
lugassawan: Addressed in 2116df2fbe1f5ca7e910d9d538fe4e29c902f9ce: added `{deps.LockfileGradleSettingsKts, "gradle-settings-kts-1", "gradle-settings-kts-2"}` as a fourth case — `TestAddWithDepsCloneGradle` now covers all four lockfile triggers end-to-end.

internal/deps/module_test.go
lugassawan: Addressed in 2116df2fbe1f5ca7e910d9d538fe4e29c902f9ce: added assertion `wantExtra := filepath.Join(subproject, DirGradleBuildOutput)` to `TestDetectModulesGradleNestedSubdir` — confirms `prefixDirs` correctly scopes ExtraDirs to the subproject path.

PR #248, opened 2026-06-03
[fix] Windows self-update: atomic binary replace (rename-aside)

### LINKED ISSUE #234: [feat] Windows self-update: atomic binary replace (rename-aside mechanism)

lugassawan:
## Summary

- Implement Windows atomic self-update via the **rename-aside** dance: `rimba.exe` → `rimba.exe.old`, copy new binary → `rimba.exe`, roll back on failure
- Sweep `rimba.exe.old` on next startup via `updater.SweepOldBinary()` wired into `Execute()`
- Core algorithm lives in untagged files (`swap_common.go`, `sweep.go`) so Linux CI fully exercises the logic; build-tagged shims (`swap_windows.go`, `sweep_windows.go`, `sweep_other.go`) provide OS dispatch
- Rollback surfaces a joined error if the restore `Rename` itself fails (binary is recoverable from `.old`)
- File created with narrow `0600` perms; `Chmod` to target perms after close — no umask-affected TOCTOU window

## Test Plan

### Type-tailored checks (fix)

- [x] `TestRenameAsideSuccess` — dst has new content, `.old` holds the old binary, perms preserved
- [x] `TestRenameAsideRollbackOnCopyError` — copy failure (dir as src → `EISDIR`) triggers rollback: dst restored, no `.old` left
- [x] `TestRenameAsideMoveAsideError` — move-aside on missing dst returns wrapped "moving aside old binary" error, no `.old` created
- [x] `TestRenameAsideStatError` — stat failure on non-existent tmpPath returns early, dst untouched
- [x] `TestCopyFileMissingSource` / `TestCopyFileDstNotCreatable` — copyFile error paths covered
- [x] `TestSweepOldBinaryRemovesOld` / `TestSweepOldBinaryNoOldFile` — sweep logic verified
- [x] `CGO_ENABLED=0 GOOS=windows go build ./...` (amd64 + arm64) — compile-checks the Windows shims without a Windows runner
- [ ] Manual: build `rimba.exe`, run `rimba update` on Windows — binary replaced, `.old` removed on next invocation

## Issue

Closes #234

### DIFF
--- a/internal/updater/swap_common.go
+++ b/internal/updater/swap_common.go
@@ -11,6 +11,8 @@ const oldBinarySuffix = ".old"
 // renameAside installs newBinary at dst using the rename-aside dance:
 // move dst → dst.old, copy newBinary → dst, chmod to match.
 // On any copy/chmod failure it rolls back by restoring dst.old → dst.
+// If the rollback rename also fails the error is surfaced so the caller
+// knows the binary is absent and the original is recoverable from dst.old.
 func renameAside(tmpPath, dst string) error {
 	info, err := os.Stat(tmpPath)
 	if err != nil {
@@ -24,8 +26,10 @@ func renameAside(tmpPath, dst string) error {
 	}
 
 	if err := copyFile(tmpPath, dst, perm); err != nil {
-		_ = os.Remove(dst)
-		_ = os.Rename(old, dst)
+		_ = os.Remove(dst) // clear any partial write; on Windows Rename fails if dst exists
+		if rbErr := os.Rename(old, dst); rbErr != nil {
+			return fmt.Errorf("installing new binary: %w; rollback failed: %w — original binary is at %s", err, rbErr, old)
+		}
 		return fmt.Errorf("installing new binary: %w", err)
 	}
 
@@ -38,7 +42,9 @@ func copyFile(src, dst string, perm os.FileMode) error {
 		return fmt.Errorf("opening source: %w", err)
 	}
 
-	out, err := os.OpenFile(dst, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, perm) //nolint:gosec // dst is the resolved current binary path, not user input
+	// Create with narrow permissions; Chmod below sets the final mode without
+	// relying on OpenFile's umask-adjusted result.
+	out, err := os.OpenFile(dst, os.O_WRONLY|os.O_CREATE|os.O_TRUNC, 0600) //nolint:gosec // dst is the resolved current binary path, not user input
 	if err != nil {
 		_ = in.Close()
 		return fmt.Errorf("creating destination: %w", err)
--- a/internal/updater/swap_common.go
+++ b/internal/updater/swap_common.go
@@ -8,11 +8,10 @@ import (
 
 const oldBinarySuffix = ".old"
 
-// renameAside installs newBinary at dst using the rename-aside dance:
-// move dst → dst.old, copy newBinary → dst, chmod to match.
-// On any copy/chmod failure it rolls back by restoring dst.old → dst.
-// If the rollback rename also fails the error is surfaced so the caller
-// knows the binary is absent and the original is recoverable from dst.old.
+// renameAside installs tmpPath at dst: moves dst → dst.old, copies tmpPath → dst,
+// then chmods to match.  Rolls back (dst.old → dst) on failure; surfaces both the
+// copy error and the rollback error if restore also fails.
+// The caller must pre-set tmpPath permissions; renameAside reads them via Stat.
 func renameAside(tmpPath, dst string) error {
 	info, err := os.Stat(tmpPath)
 	if err != nil {
@@ -26,9 +25,9 @@ func renameAside(tmpPath, dst string) error {
 	}
 
 	if err := copyFile(tmpPath, dst, perm); err != nil {
-		_ = os.Remove(dst) // clear any partial write; on Windows Rename fails if dst exists
+		rmErr := os.Remove(dst) // clear any partial write; on Windows Rename fails if dst exists
 		if rbErr := os.Rename(old, dst); rbErr != nil {
-			return fmt.Errorf("installing new binary: %w; rollback failed: %w — original binary is at %s", err, rbErr, old)
+			return fmt.Errorf("installing new binary: %w; rollback failed (remove: %w, rename: %w) — original binary is at %s", err, rmErr, rbErr, old)
 		}
 		return fmt.Errorf("installing new binary: %w", err)
 	}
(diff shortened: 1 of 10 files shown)


internal/updater/swap_common.go
```diff
@@ -0,0 +1,69 @@
...
+	}
+	perm := info.Mode().Perm()
+
+	old := dst + oldBinarySuffix
+	if err := os.Rename(dst, old); err != nil {
+		return fmt.Errorf("moving aside old binary: %w", err)
+	}
+
+	if err := copyFile(tmpPath, dst, perm); err != nil {
+		_ = os.Remove(dst) // clear any partial write; on Windows Rename fails if dst exists
```
lugassawan: **Medium:** `os.Remove(dst)` error is silently discarded before the rollback rename, hiding the root cause when rollback also fails.

On Windows, if the partial write of `dst` leaves the file locked and `os.Remove(dst)` fails, the subsequent `os.Rename(old, dst)` also fails (Windows cannot rename over an existing path). The caller sees the compound "rollback failed" message but loses the underlying `Remove` error — the most actionable diagnosis on the platform this feature targets.

**Suggested fix:** capture and fold the remove error into the compound message:
```go
if rmErr := os.Remove(dst); rmErr != nil {
    return fmt.Errorf("installing new binary: %w; rollback failed (remove: %v, rename: %w) — original binary is at %s", err, rmErr, rbErr, old)
}
```

internal/updater/swap_common_test.go
```diff
@@ -0,0 +1,170 @@
...
+	info, err := os.Stat(dst)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if info.Mode().Perm() != 0755 {
+		t.Errorf("dst perm = %04o, want 0755", info.Mode().Perm())
+	}
+}
+
+func TestRenameAsideRollbackOnCopyError(t *testing.T) {
```
lugassawan, line 53: **Medium:** The double-failure path — copy fails AND rollback rename fails — has zero test coverage.

The error string `"installing new binary: %w; rollback failed: %w — original binary is at %s"` (swap_common.go:31) is the most critical recovery hint in the feature (binary absent, original in `.old`), yet it is never exercised by the test suite.

**Suggested fix:** add a test that forces both failures — e.g., after the move-aside completes, chmod the parent directory to 0555 so neither `OpenFile(dst)` nor `Rename(old, dst)` can succeed — and assert the returned error contains both `"rollback failed"` and the `.old` path.

internal/updater/sweep_test.go
```diff
@@ -0,0 +1,36 @@
...
+
+func TestSweepOldBinaryNoOldFile(t *testing.T) {
+	tmpDir := t.TempDir()
+	exe := filepath.Join(tmpDir, "rimba")
+
+	// No exe.old present — best-effort sweep should not panic.
+	sweepOldBinary(exe)
+}
+
+func TestSweepOldBinaryExported(t *testing.T) {
```
lugassawan: **Low:** `TestSweepOldBinaryExported` always calls the non-Windows no-op (`sweep_other.go`); the name implies broader coverage than it provides.

The `os.Executable`/`filepath.EvalSymlinks` paths in `sweep_windows.go` are compile-checked but never run. A reader skimming the test names will assume the exported function is exercised end-to-end.

**Suggested fix:** rename to `TestSweepOldBinaryNonWindowsNoOp` and add a comment: `// sweep_windows.go is compile-checked by the build-cross CI job; runtime coverage is Windows-only.`

internal/updater/swap_common.go
```diff
@@ -0,0 +1,69 @@
...
+
+import (
+	"fmt"
+	"io"
+	"os"
+)
+
+const oldBinarySuffix = ".old"
+
+// renameAside installs newBinary at dst using the rename-aside dance:
```
lugassawan: **Nitpick:** The doc comment does not mention the implicit caller contract: `tmpPath` must be pre-chmod'd to the desired final permissions before calling.

`copyFile` derives `perm` by stat-ing `tmpPath`, so callers that skip the chmod (unlike `Replace()`'s `tmp.Chmod` at updater.go:257) will silently install the binary with wrong permissions.

**Suggested fix:** add one sentence: *"The caller must set the desired final permissions on tmpPath before calling; renameAside reads them via Stat and applies them to the installed binary."*

internal/updater/swap_common.go
lugassawan: Addressed in ade1cc6: surface rmErr via %w in the compound rollback message so all three errors (copy, remove, rename) are wrapped and inspectable via errors.Is.

internal/updater/swap_common_test.go
lugassawan: Addressed in ade1cc6: added TestRenameAsideRollbackDoubleFailSubcomponents — manually performs the move-aside, chmodds the parent dir to 0555, then confirms both copyFile and the rollback Rename fail in that context. The compound fmt.Errorf line itself requires mid-function interception to reach through renameAside; the sub-component test validates the conditions that feed into it.

internal/updater/sweep_test.go
lugassawan: Addressed in ade1cc6: renamed to TestSweepOldBinaryNonWindowsNoOp with a comment directing readers to sweep_windows.go and the build-cross CI job.

internal/updater/swap_common.go
lugassawan: Addressed in ade1cc6: simplified doc comment and added 'The caller must pre-set tmpPath permissions; renameAside reads them via Stat.'

PR #256, opened 2026-06-04
[fix] Protect worktree operands with separator

loopassembly:
Closes #252.

## What changed
- Pass `AddWorktree` path/source operands after `--`, matching the existing `AddWorktreeFromBranch` hardening.
- Build `MoveWorktree` args with force flags first, then `--`, then the old/new path operands.
- Add focused mock-runner tests for the exact git argv shape.

## Why
This keeps dash-leading sources or paths from being parsed as git options, while preserving the existing double-`--force` behavior for locked worktree moves.

## Validation
- `go test ./internal/git`
- `go test ./...`

### DIFF
--- a/internal/git/mock_runner_test.go
+++ b/internal/git/mock_runner_test.go
@@ -349,6 +349,25 @@ func TestMoveWorktreeNoForce(t *testing.T) {
 	}
 }
 
+func TestMoveWorktreeInsertsDashDash(t *testing.T) {
+	var captured []string
+	r := &mockRunner{
+		run: func(args ...string) (string, error) {
+			captured = args
+			return "", nil
+		},
+	}
+
+	if err := MoveWorktree(r, "/old/path", "/new/path", false); err != nil {
+		t.Fatalf("MoveWorktree: %v", err)
+	}
+
+	want := []string{cmdWorktree, "move", "--", "/old/path", "/new/path"}
+	if !slices.Equal(captured, want) {
+		t.Errorf("args = %v, want %v", captured, want)
+	}
+}
+
 func TestMoveWorktreeForce(t *testing.T) {
 	var captured []string
 	r := &mockRunner{
@@ -362,14 +381,9 @@ func TestMoveWorktreeForce(t *testing.T) {
 		t.Fatalf("MoveWorktree: %v", err)
 	}
 	// git worktree move requires --force twice to move locked worktrees
-	forceCount := 0
-	for _, a := range captured {
-		if a == flagForce {
-			forceCount++
-		}
-	}
-	if forceCount != 2 {
-		t.Errorf("expected 2 %s flags, got %d in args %v", flagForce, forceCount, captured)
+	want := []string{cmdWorktree, "move", flagForce, flagForce, "--", "/old/path", "/new/path"}
+	if !slices.Equal(captured, want) {
+		t.Errorf("args = %v, want %v", captured, want)
 	}
 }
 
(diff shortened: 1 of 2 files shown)


internal/git/worktree.go
```diff
@@ -50,10 +50,11 @@ func RemoveWorktree(r Runner, path string, force bool) error {
 // MoveWorktree moves the worktree from oldPath to newPath.
 // When force is true, --force is passed twice so that even locked worktrees can be moved.
 func MoveWorktree(r Runner, oldPath, newPath string, force bool) error {
-	args := []string{cmdWorktree, "move", oldPath, newPath}
+	args := []string{cmdWorktree, "move"}
 	if force {
 		args = append(args, flagForce, flagForce)
 	}
+	args = append(args, "--", oldPath, newPath)
```
lugassawan, line 59: **High — `git worktree move` may not support `--` as end-of-options separator**

`git worktree add` explicitly lists `--` in its synopsis; `git worktree move` does not. The generated invocation is now:

```
git worktree move --force --force -- /old /new
```

If `move` does not recognise `--` as end-of-options, git reads `--` as `<worktree>` and `/old` as `<new-path>`, silently moving the wrong worktree (or erroring). Verify empirically in a temp repo:

```sh
git worktree move -- /some/path /other/path
```

If git rejects it, the `--` must be omitted for `move` (or conditioned on the operands actually starting with `-`). A targeted integration test would catch this definitively.

internal/git/mock_runner_test.go
```diff
@@ -349,6 +349,25 @@ func TestMoveWorktreeNoForce(t *testing.T) {
 	}
 }
 
+func TestMoveWorktreeInsertsDashDash(t *testing.T) {
```
lugassawan, line 328: **Medium — `TestMoveWorktreeNoForce` is now fully subsumed by this test**

`TestMoveWorktreeNoForce` (line 335) calls `MoveWorktree(r, "/old/path", "/new/path", false)` and asserts `!slices.Contains(captured, flagForce)`. This new test makes the same call and asserts `slices.Equal(captured, want)` — which already guarantees no `--force` (since `want` excludes it). The older test adds zero additional coverage.

Suggested: delete `TestMoveWorktreeNoForce` (or collapse both no-force tests into a table-driven test alongside `TestMoveWorktreeForce` to keep all three cases in one place).

lugassawan (commented):
**Review Decision: COMMENT**

---
_Reviewed by `reviewer` ([swe-workbench](https://github.com/lugassawan/swe-workbench)). Posted 2 inline comments, deduped 0._

**Medium — `RemoveWorktree` (worktree.go:42) still passes `path` as a bare positional (not in this diff)**

`RemoveWorktree` was not in the #229 hardening sweep and was not touched in this PR either. A `path` beginning with `-` is passed directly to git and parsed as a flag. Suggested: restructure to `["worktree", "remove", "--force"?, "--", path]`, mirroring the pattern established by this PR — and empirically verify that `git worktree remove` honours `--` (same check recommended for `move`).

internal/git/worktree.go
lugassawan: Verified: `git worktree move` does support `--` as an end-of-options separator (confirmed empirically on git 2.55.0 — `git worktree move --force --force -- ../wt1 ../wt2` succeeds against a real worktree in a temp repo, and paths are treated as positional operands, not misread as `<worktree>`/`<new-path>`). The existing real-git integration test `TestMoveWorktree` in internal/git/worktree_test.go now transits this exact `--`-protected invocation and passes, giving the targeted integration coverage requested here. Addressed in 622fe9f.

internal/git/mock_runner_test.go
lugassawan: Agreed — deleted `TestMoveWorktreeNoForce`. `TestMoveWorktreeInsertsDashDash` (exact-argv `slices.Equal` assertion) already implies the no-`--force` case since `want` excludes it, so the older test added no coverage. Addressed in 622fe9f.

PR #258, opened 2026-06-05
[fix] clean --force force-removes dirty merged/stale worktrees

lugassawan:
## Summary
- Thread `force bool` end-to-end through `removeAndCleanup` and `RemoveCandidates` so `clean --force` actually passes `--force` to `git worktree remove`
- Previously `--force` only skipped the confirmation prompt; dirty worktrees (e.g. with `node_modules/`) were silently left in place
- MCP `clean` tool now always force-removes candidates (unattended, no prompt to skip)
- Update `--force` flag description and hint text to document the new behavior

## Test Plan
- [x] New unit tests: `TestRemoveAndCleanupForceFlag`, `TestRemoveCandidatesForceFlag` — assert `--force` is/isn't in git worktree args
- [x] New e2e test: `TestCleanMergedForceDirtyWorktree` — verifies `clean --merged --force` removes a worktree with uncommitted staged changes
- [x] Manual repro: `clean --merged` (without `--force`) refused dirty worktree; `clean --merged --force` succeeded

## Issue

Closes #257

### DIFF
--- a/internal/mcp/tool_clean.go
+++ b/internal/mcp/tool_clean.go
@@ -117,7 +117,7 @@ func mcpCleanMerged(ctx context.Context, r git.Runner, hctx *HandlerContext, dry
 	// Force mode: no confirmation prompts. Probe origin once and pass the result
 	// directly so RemoveCandidates does not re-issue git remote get-url per candidate.
 	originPresent := git.RemoteExists(r, git.DefaultRemote)
-	opItems := operations.RemoveCandidates(ctx, r, mergedResult.Candidates, originPresent, nil)
+	opItems := operations.RemoveCandidates(ctx, r, mergedResult.Candidates, originPresent, true, nil)
 	warnings := mergedResult.Warnings
 	items := make([]cleanedItem, len(opItems))
 	for i, item := range opItems {
@@ -163,7 +163,7 @@ func mcpCleanStale(ctx context.Context, r git.Runner, hctx *HandlerContext, dryR
 		toRemove[i] = c.CleanCandidate
 	}
 
-	opItems := operations.RemoveCandidates(ctx, r, toRemove, false, nil)
+	opItems := operations.RemoveCandidates(ctx, r, toRemove, false, true, nil)
 	items := make([]cleanedItem, len(opItems))
 	for i, item := range opItems {
 		items[i] = cleanedItem{Branch: item.Branch, Path: item.Path}
--- a/internal/mcp/tool_clean.go
+++ b/internal/mcp/tool_clean.go
@@ -21,6 +21,9 @@ func registerCleanTool(s *server.MCPServer, hctx *HandlerContext) {
 		mcp.WithBoolean("dry_run",
 			mcp.Description("Preview what would be cleaned without making changes"),
 		),
+		mcp.WithBoolean("force",
+			mcp.Description("Force-remove dirty worktrees even if they contain uncommitted changes (used with mode=merged or mode=stale)"),
+		),
 		mcp.WithNumber("stale_days",
 			mcp.Description("Number of days to consider a worktree stale (default: 14, used with mode=stale)"),
 		),
@@ -36,6 +39,7 @@ func handleClean(hctx *HandlerContext) server.ToolHandlerFunc {
 		}
 
 		dryRun := req.GetBool("dry_run", false)
+		force := req.GetBool("force", false)
 		staleDays := req.GetInt("stale_days", 14)
 
 		r := hctx.Runner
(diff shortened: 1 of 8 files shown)


internal/mcp/tool_clean.go
```diff
@@ -117,7 +117,7 @@ func mcpCleanMerged(ctx context.Context, r git.Runner, hctx *HandlerContext, dry
 	// Force mode: no confirmation prompts. Probe origin once and pass the result
 	// directly so RemoveCandidates does not re-issue git remote get-url per candidate.
 	originPresent := git.RemoteExists(r, git.DefaultRemote)
-	opItems := operations.RemoveCandidates(ctx, r, mergedResult.Candidates, originPresent, nil)
+	opItems := operations.RemoveCandidates(ctx, r, mergedResult.Candidates, originPresent, true, nil)
```
lugassawan: **High: MCP `clean` tool silently hardcodes `force=true`, permanently destroying dirty worktrees with no caller opt-out**

Before this PR the tool implicitly used `force=false`; this PR upgrades both `mcpCleanMerged` and `mcpCleanStale` to `force=true` unconditionally. The MCP schema exposes `dry_run` but no `force` parameter — an MCP client calling `clean` with `mode=merged` on a worktree with untracked build artefacts or local config edits will silently force-remove it, with no schema parameter to opt out.

The comment on line 117 reads "Force mode: no confirmation prompts" — accurate for the CLI, but irrelevant for the MCP tool (there is no confirmation prompt anyway). The behaviour change is non-additive: callers lose the previous safe default with no visibility.

**Suggested fix:** expose a `force` boolean in `registerCleanTool` (defaulting to `false`) and thread it through `handleClean` → `mcpCleanMerged`/`mcpCleanStale`:

```go
// registerCleanTool — alongside dry_run:
mcp.WithBoolean("force",
    mcp.Description("Force-remove dirty worktrees even if they contain uncommitted changes"),
),

// handleClean:
force := req.GetBool("force", false)
// pass force down to mcpCleanMerged/mcpCleanStale and into RemoveCandidates
```

tests/e2e/clean_test.go
```diff
@@ -141,6 +141,25 @@ func TestCleanMergedForce(t *testing.T) {
...
+	wtPath := cleanMergeSetup(t, repo, "dirty-force")
+
+	// Dirty the worktree with a staged tracked file so git worktree remove
+	// without --force would refuse.
+	testutil.CreateFile(t, wtPath, "uncommitted.txt", "dirty content")
+	testutil.GitCmd(t, wtPath, "add", "uncommitted.txt")
+
+	r := rimbaSuccess(t, repo, "clean", flagMergedE2E, flagForceE2E)
+	assertContains(t, r.Stdout, msgRemovedWorktree)
+	assertContains(t, r.Stdout, "Cleaned 1 merged worktree(s)")
```
lugassawan, line 160: **Medium: `TestCleanMergedForceDirtyWorktree` does not assert branch deletion**

The test verifies the worktree directory is gone (`assertFileNotExists`) and the success summary is printed, but unlike its clean-path counterpart `TestCleanMergedForce` (line 130–131), it does not assert `msgDeletedBranch`. A regression where the branch is left behind after force-removal would go undetected.

**Suggested fix:** add the branch-deletion assertion:

```go
assertContains(t, r.Stdout, msgRemovedWorktree)
assertContains(t, r.Stdout, msgDeletedBranch)   // add this
assertContains(t, r.Stdout, "Cleaned 1 merged worktree(s)")
```

internal/operations/clean_test.go
```diff
@@ -466,3 +466,35 @@ func TestRemoveCandidatesNoOriginSkipsRemoteDelete(t *testing.T) {
...
+						capturedArgs = args
+					}
+					return "", nil
+				},
+				runInDir: noopRunInDir,
+			}
+			candidates := []CleanCandidate{{Path: "/wt/a", Branch: "feature/a"}}
+			RemoveCandidates(context.Background(), r, candidates, false, tt.force, nil)
+			joined := strings.Join(capturedArgs, " ")
+			hasForce := strings.Contains(joined, "--force")
```
lugassawan: **Low: `--force` detection via joined-string substring match is slightly fragile**

`strings.Contains(joined, "--force")` would match a hypothetical future `--force-with-lease` or similar flag. The slice-element check is more precise at no extra cost:

```go
// instead of:
joined := strings.Join(capturedArgs, " ")
hasForce := strings.Contains(joined, "--force")

// prefer:
hasForce := slices.Contains(capturedArgs, "--force")
```

internal/mcp/tool_clean.go
lugassawan: Addressed in 7d7bc9b: exposed `force bool` parameter in MCP clean tool schema (defaults to `false`), threaded through `handleClean` → `mcpCleanMerged`/`mcpCleanStale` → `RemoveCandidates`. Removed the hardcoded `true`.

tests/e2e/clean_test.go
lugassawan: Addressed in 7d7bc9b: added `assertContains(t, r.Stdout, msgDeletedBranch)` to `TestCleanMergedForceDirtyWorktree`.

internal/operations/clean_test.go
lugassawan: Addressed in 7d7bc9b: switched to `slices.Contains(capturedArgs, "--force")` in both `TestRemoveCandidatesForceFlag` and `TestRemoveAndCleanupForceFlag`.

PR #259, opened 2026-06-05
[fix] copy_files cannot escape worktree via symlinked dst component

lugassawan:
## Summary
- `copy_files` (via `fileutil.CopyEntries`) now detects when a destination path component is a symlink pointing outside the worktree and rejects the write with `ErrPathEscapes`
- Added `resolveDstRoot`, `resolveExistingAncestor`, and `assertContained` helpers in `internal/fileutil/copy.go`; all three write syscalls (`MkdirAll` in `copyEntry`, `MkdirAll` in `copyDir`, `OpenFile` in `copyFile`) are now guarded
- `ContainedJoin` (lexical `..` guard) is unchanged; the new check adds symlink-resolution on top of it
- When `dst` doesn't exist at call time (tests with mocked runners, or pre-worktree-creation), the check is a no-op — in production `git worktree add` always creates `dst` before `CopyEntries` runs

## Test Plan
- [x] Unit tests pass (`make test`) — 1990 passed
- [x] E2E tests pass (`make test-e2e`) — 301 passed
- [x] Coverage at 97.0% (`make test-coverage`)
- New tests added: `TestCopyEntriesRejectsSymlinkedDstEscape`, `TestCopyEntriesRejectsNestedSymlinkEscape`, `TestCopyEntriesBenignSymlinkInsideDst`, `TestCopyEntriesNonExistentDst` (updated), `TestCopyEntriesDstResolvePermissionError`, `TestCopyEntriesAncestorPermissionError`

## Issue

Closes #254

### DIFF
--- a/internal/fileutil/copy_test.go
+++ b/internal/fileutil/copy_test.go
@@ -655,3 +655,144 @@ func TestCopyEntriesRecursiveCopyDirError(t *testing.T) {
 		t.Errorf(errContainsFmt, err, "copy "+dotConfig+":")
 	}
 }
+
+func TestCopyEntriesRejectsSymlinkedDstEscape(t *testing.T) {
+	src := t.TempDir()
+	dst := t.TempDir()
+	outside := t.TempDir()
+
+	if err := os.Symlink(outside, filepath.Join(dst, "sub")); err != nil {
+		t.Fatal(err)
+	}
+	_ = os.MkdirAll(filepath.Join(src, "sub"), 0755)
+	_ = os.WriteFile(filepath.Join(src, "sub", "secret.txt"), []byte("secret"), 0644)
+
+	_, _, err := fileutil.CopyEntries(src, dst, []string{"sub/secret.txt"})
+	if err == nil {
+		t.Fatal("expected error when dst has top-level symlink pointing outside")
+	}
+	if !errors.Is(err, fileutil.ErrPathEscapes) {
+		t.Fatalf("error %v does not wrap fileutil.ErrPathEscapes", err)
+	}
+	if !strings.Contains(err.Error(), "resolves outside") {
+		t.Errorf("error %q should mention resolves outside", err.Error())
+	}
+	if _, statErr := os.Stat(filepath.Join(outside, "secret.txt")); !os.IsNotExist(statErr) {
+		t.Error("secret.txt must not have been written to outside")
+	}
+}
+
+func TestCopyEntriesRejectsNestedSymlinkEscape(t *testing.T) {
+	src := t.TempDir()
+	dst := t.TempDir()
+	outside := t.TempDir()
+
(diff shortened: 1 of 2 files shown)


internal/fileutil/copy_test.go
```diff
@@ -655,3 +655,144 @@ func TestCopyEntriesRecursiveCopyDirError(t *testing.T) {
...
+	}
+	if !strings.Contains(err.Error(), "resolves outside") {
+		t.Errorf("error %q should mention resolves outside", err.Error())
+	}
+	if _, statErr := os.Stat(filepath.Join(outside, "secret.txt")); !os.IsNotExist(statErr) {
+		t.Error("secret.txt must not have been written to outside")
+	}
+}
+
+func TestCopyEntriesRejectsNestedSymlinkEscape(t *testing.T) {
```
lugassawan, line 685: **Missing test: top-level directory entry where `dst/<name>` is itself an outbound symlink**

`TestCopyEntriesRejectsNestedSymlinkEscape` covers the case where a *sub-directory inside the top-level entry* is a symlink (`dst/d/deep -> outside`). There is no test for the case where the *top-level destination directory itself* is a symlink (`dst/mydir -> outside`) and `mydir` is the copied entry.

The code handles it correctly — `copyDir` calls `assertContained(dstRoot, dst)` before `MkdirAll` — but if that guard were accidentally removed in a future refactor, this scenario would regress silently.

**Suggested addition:**
```go
func TestCopyEntriesRejectsTopLevelDirSymlinkEscape(t *testing.T) {
    src := t.TempDir()
    dst := t.TempDir()
    outside := t.TempDir()

    _ = os.Mkdir(filepath.Join(src, "mydir"), 0755)
    _ = os.WriteFile(filepath.Join(src, "mydir", "f.txt"), []byte("data"), 0644)
    if err := os.Symlink(outside, filepath.Join(dst, "mydir")); err != nil {
        t.Fatal(err)
    }

    _, _, err := fileutil.CopyEntries(src, dst, []string{"mydir"})
    if err == nil {
        t.Fatal("expected error when dst top-level entry is a symlink pointing outside")
    }
    if !errors.Is(err, fileutil.ErrPathEscapes) {
        t.Fatalf("error %v does not wrap fileutil.ErrPathEscapes", err)
    }
    entries, _ := os.ReadDir(outside)
    if len(entries) != 0 {
        t.Errorf("outside dir should be empty, got %v", entries)
    }
}
```

lugassawan: Addressed in 31441d5: added `TestCopyEntriesRejectsTopLevelDirSymlinkEscape` — creates `dst/mydir -> outside` symlink, copies entry `"mydir"`, asserts `errors.Is(err, ErrPathEscapes)` and outside dir stays empty.

PR #260, opened 2026-06-05
[feat] honor global --json flag in log

lugassawan:
## Summary

- Add `LogItem` DTO to `internal/output/types.go` with fields: `task`, `service` (omitempty), `type`, `branch`, `path`, `last_commit` (RFC3339 UTC), `subject` — mirroring `ListItem` conventions
- Thread `path` through `logEntry` struct from `git.WorktreeEntry.Path` (was missing, needed for JSON output)
- Add JSON guards at all three exit paths in `logCmd.RunE` (no worktrees, no commits after filters, happy path)
- Add `writeLogJSON` helper that maps `[]logEntry → []output.LogItem` and emits via `output.WriteJSON`
- `--since` / `--limit` filtering is applied before the JSON branch, so both flags are honoured in JSON mode
- Spinner already suppressed in JSON mode via existing `spinnerOpts(cmd)` — no change needed
- **Audit:** `status` and `list` already implement `--json` with passing tests. One asymmetry noted: `StatusItem` lacks `path` (present in `ListItem`/`ExecResult`/`LogItem`); left as-is — additive change deferred to avoid schema churn on an established shape

## Test Plan

- [x] New `TestLogJSON` unit test: happy path with field assertions (`task`, `type`, `branch`, `path`, `last_commit`, `subject`)
- [x] New `TestLogJSONEmpty` unit test: no-worktrees path emits `{"data":[]}`
- [x] New `TestLogJSON` e2e test: real git repo, ANSI-free stdout, empty stderr

## Issue

Closes #255

### DIFF
--- a/cmd/log.go
+++ b/cmd/log.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/lugassawan/rimba/internal/git"
+	"github.com/lugassawan/rimba/internal/output"
 	"github.com/lugassawan/rimba/internal/parallel"
 	"github.com/lugassawan/rimba/internal/resolver"
 	"github.com/lugassawan/rimba/internal/spinner"
@@ -25,6 +26,7 @@ type logEntry struct {
 	task       string
 	service    string
 	typeName   string
+	path       string
 	commitTime time.Time
 	subject    string
 	valid      bool
@@ -53,6 +55,9 @@ var logCmd = &cobra.Command{
 		candidates := git.FilterEntries(entries, mainBranch)
 
 		if len(candidates) == 0 {
+			if isJSON(cmd) {
+				return output.WriteJSON(cmd.OutOrStdout(), version, "log", make([]output.LogItem, 0))
+			}
 			fmt.Fprintln(cmd.OutOrStdout(), "No worktrees found.")
 			return nil
 		}
@@ -88,8 +93,15 @@ var logCmd = &cobra.Command{
 		}
 
 		if len(valid) == 0 {
+			if isJSON(cmd) {
+				return output.WriteJSON(cmd.OutOrStdout(), version, "log", make([]output.LogItem, 0))
+			}
 			fmt.Fprintln(cmd.OutOrStdout(), "No recent commits found.")
 			return nil
+		}
+
+		if isJSON(cmd) {
+			return writeLogJSON(cmd, valid)
 		}
 
 		noColor, _ := cmd.Flags().GetBool(flagNoColor)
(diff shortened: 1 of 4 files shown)


cmd/log.go
```diff
@@ -53,6 +55,9 @@ var logCmd = &cobra.Command{
 		candidates := git.FilterEntries(entries, mainBranch)
 
 		if len(candidates) == 0 {
+			if isJSON(cmd) {
+				return output.WriteJSON(cmd.OutOrStdout(), version, "log", make([]output.LogItem, 0))
```
lugassawan: **Medium:** Both early-exit JSON paths call `make([]output.LogItem, 0)` inline, which duplicates allocation logic. The `list` command addresses this with a `listRenderEmpty` helper. Consider extracting a `writeLogJSONEmpty` helper to keep the pattern consistent:

```go
func writeLogJSONEmpty(cmd *cobra.Command) error {
    return output.WriteJSON(cmd.OutOrStdout(), version, "log", make([]output.LogItem, 0))
}
```

No behavioral bug — purely a consistency observation.

tests/e2e/json_test.go
```diff
@@ -112,6 +116,66 @@ func TestListJSONArchived(t *testing.T) {
...
+
+	if env["command"] != "log" {
+		t.Errorf("command = %v, want 'log'", env["command"])
+	}
+
+	data, ok := env["data"].([]any)
+	if !ok {
+		t.Fatalf("data type = %T, want []any", env["data"])
+	}
+	if len(data) == 0 {
```
lugassawan, line 147: **Low:** `data[0]` is assumed to be the `log-json-task` entry, but in a repo with multiple worktrees the ordering of JSON output is commit-time-based, and the first item may not be the one under test. Consider searching the slice for the expected task instead of indexing directly:

```go
var found map[string]any
for _, d := range data {
    if m, ok := d.(map[string]any); ok && m["task"] == "log-json-task" {
        found = m
        break
    }
}
if found == nil {
    t.Fatal("log-json-task entry not found in JSON output")
}
```

Not a current failure risk (only one worktree is created in this test), but makes the test robust to future `setupInitializedRepo` changes.

cmd/log.go
lugassawan: Addressed in 5bb58d66f8b30c07312560a3e34324b0cf5cedba: extracted `writeLogJSONEmpty` helper — both early-exit paths now call `writeLogJSONEmpty(cmd)` instead of inlining `output.WriteJSON(..., make([]output.LogItem, 0))`.

tests/e2e/json_test.go
lugassawan: Addressed in 5bb58d66f8b30c07312560a3e34324b0cf5cedba: replaced `data[0]` indexing with a loop that searches for the `log-json-task` entry by task name — now order-independent.

cmd/log_test.go
```diff
@@ -262,3 +265,138 @@ func TestLogCommitInfoError(t *testing.T) {
...
+
+	data, ok := env.Data.([]any)
+	if !ok {
+		t.Fatalf("data type = %T, want []any", env.Data)
+	}
+	if len(data) != 1 {
+		t.Errorf("data length = %d, want 1", len(data))
+	}
+
+	item, ok := data[0].(map[string]any)
```
lugassawan: **Low:** The unit test still accesses `data[0]` directly after the e2e test was updated to search by task name. The `len(data) != 1` guard at line 297 makes this safe today (only one mock worktree), but it diverges from the search-by-task pattern the PR intentionally established in the e2e test.

Consider adding a comment documenting the single-entry precondition, or applying the same search-by-task idiom for consistency:

```go
var item map[string]any
for _, d := range data {
    if m, ok := d.(map[string]any); ok && m["task"] == taskLogin {
        item = m
        break
    }
}
if item == nil {
    t.Fatalf("entry for %q not found in JSON output", taskLogin)
}
```

PR #261, opened 2026-06-05
[fix] Force stable locale for git commands

ibobgunardi:
Fixes #251.

ExecRunner was inheriting the caller's locale, but branch cleanup treats a few already-gone git errors as success by matching git's English output. A translated git message can miss those checks and turn cleanup into a hard error.

This strips inherited LANG and LC_ALL values for git subprocesses and sets both to C so git output stays stable.

Tested on Windows:
```sh
go test ./internal/git -run "TestStableGitEnvForcesCLocale|TestRun" -count=1 -v
go test ./internal/git -run "Test(Run|StableGitEnv|DeleteBranch|DeleteRemoteBranch)" -count=1 -v
```

### DIFF
--- a/internal/git/git.go
+++ b/internal/git/git.go
@@ -3,6 +3,7 @@ package git
 import (
 	"context"
 	"fmt"
+	"os"
 	"os/exec"
 	"strings"
 )
@@ -24,6 +25,7 @@ type ExecRunner struct {
 // RunInDirContext is the single execution primitive: all other methods delegate here.
 func (r *ExecRunner) RunInDirContext(ctx context.Context, dir string, args ...string) (string, error) {
 	cmd := exec.CommandContext(ctx, "git", args...)
+	cmd.Env = stableGitEnv(os.Environ())
 	if dir != "" {
 		cmd.Dir = dir
 	}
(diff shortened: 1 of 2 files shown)


lugassawan (commented):
Overall the locale-forcing approach is correct and well-placed — the env-stripping logic belongs exactly here. One coverage issue is blocking CI; see inline.

internal/git/git.go
```diff
@@ -36,6 +38,23 @@ func (r *ExecRunner) RunInDirContext(ctx context.Context, dir string, args ...st
 	return result, nil
 }
 
+func stableGitEnv(environ []string) []string {
+	env := make([]string, 0, len(environ)+2)
```
lugassawan, line 42: This `!ok` branch (a `=`-less env entry) is unreachable in practice: `os.Environ()` always returns `KEY=VALUE` strings. Because the test does not exercise it, total coverage drops to 96.9% — below the 97% gate — and the `coverage` CI job is currently failing.

The branch is also redundant: a `=`-less entry would have `key = entry`, fail both `EqualFold` checks, and reach the same `append` below, so deleting the block is behaviourally identical.

Simplest fix — delete the `if !ok` block:

```go
func stableGitEnv(environ []string) []string {
	env := make([]string, 0, len(environ)+2)
	for _, entry := range environ {
		key, _, ok := strings.Cut(entry, "=")
		if !ok {
			env = append(env, entry)
			continue
		}                         // ← delete lines 42-45
		if strings.EqualFold(key, "LANG") || strings.EqualFold(key, "LC_ALL") {
			continue
		}
		env = append(env, entry)
	}
	return append(env, "LANG=C", "LC_ALL=C")
}
```

Alternative: keep the branch and add a `=`-less input to `TestStableGitEnvForcesCLocale` (e.g. `"NOEQUALS"`) so coverage is exercised — though the branch is still dead code in production.

lugassawan (approved):
**Review Decision: APPROVE**

---
_Re-reviewed by `reviewer` ([swe-workbench](https://github.com/lugassawan/swe-workbench)). Posted 0 inline comments, deduped 0._

lugassawan (approved):
**Review Decision: APPROVE**

---
_Re-reviewed by `reviewer` ([swe-workbench](https://github.com/lugassawan/swe-workbench)). Posted 0 inline comments, deduped 0._

PR #262, opened 2026-06-05
[fix] Handle same-name worktree rename

### LINKED ISSUE #220: [bug] rimba rename to the same name errors with a hint that deletes your own branch
## Product framing

**Problem:** A no-op `rimba rename <task> <same-task>` returns a confusing "branch already exists"
error and suggests `git branch -D <branch>`, which deletes the user's own current branch.

**Value:** New contributors who accidentally rename to the same name get a safe, clear message
instead of a footgun.

**Acceptance criteria:**
- Given a worktree on branch X, when `rimba rename` resolves to the same branch X, the error reads
  like "new name is the same as the current name" with no destructive `git branch -D` hint.
- A test covers the same-name path.

**Impact / Effort:** Impact: S. Effort: S — one early guard + one test, single file.

## Surface affected

- [x] Command (`cmd/*`) — `cmd/rename.go`
- [x] Worktree operations / hooks (`internal/operations/`) — `internal/operations/rename.go`

**Specific surface:** `internal/operations/rename.go:31` (the `git.BranchExists` guard fires for a
same-name rename).

## Environment

- OS: macOS
- Shell: zsh
- git version:

## Reproduction steps

1. `rimba add auth` (creates `feature/auth` + worktree).
2. `rimba rename auth auth`.
3. Observe: `branch "feature/auth" already exists` with hint `git branch -D feature/auth`.

## Expected behavior

A clear error such as "new name is the same as the current name" — and no hint that would delete
the current branch.

## Actual behavior

Confusing "already exists" error plus a destructive `git branch -D` suggestion.

## Logs / screenshots (optional)

> **Good first issue:** add an early guard in `RenameWorktree`
> (`internal/operations/rename.go`, before the `BranchExists` check at line 31):
> if `newBranch == wt.Branch`, return a friendly "same name" error. Add a same-name test case
> alongside the existing rename tests.

ibobgunardi:
Fixes #220.

`rename` now catches the no-op case before checking whether the target branch exists. That keeps `rimba rename auth auth` from returning the generic "branch already exists" error and avoids showing the `git branch -D` recovery hint for the current branch.

Tested on Windows:
```sh
go test ./internal/operations -run TestRenameWorktreeSameName -count=1 -v
go test ./internal/operations -run TestRenameWorktree -count=1 -v
```

I also tried `go test ./internal/operations -count=1`, but this local Windows environment fails existing tests that need symlink privileges and `sh` for hook execution.

### DIFF
--- a/internal/operations/rename.go
+++ b/internal/operations/rename.go
@@ -27,6 +27,9 @@ func RenameWorktree(r git.Runner, wt resolver.WorktreeInfo, newTask, wtDir strin
 	}
 
 	newBranch := resolver.FullBranchName(svc, matchedPrefix, newTask)
+	if newBranch == wt.Branch {
+		return RenameResult{}, fmt.Errorf("new name is the same as the current name: %q", newTask)
+	}
 
 	if git.BranchExists(r, newBranch) {
 		return RenameResult{}, errhint.WithFix(
--- a/internal/operations/rename_test.go
+++ b/internal/operations/rename_test.go
@@ -72,6 +72,38 @@ func TestRenameWorktreeBranchExists(t *testing.T) {
 	}
 }
 
+func TestRenameWorktreeSameName(t *testing.T) {
+	calls := 0
+	r := &mockRunner{
+		run: func(args ...string) (string, error) {
+			calls++
+			return "", nil
+		},
+		runInDir: func(dir string, args ...string) (string, error) {
+			calls++
+			return "", nil
+		},
+	}
+
+	wt := resolver.WorktreeInfo{Branch: branchFeature, Path: pathWtFeatureLogin}
+	_, err := RenameWorktree(r, wt, "login", wtDir, false)
+	if err == nil {
+		t.Fatal("expected error for same-name rename")
+	}
+	if !strings.Contains(err.Error(), "same as the current name") {
+		t.Errorf("error = %q, want same-name message", err.Error())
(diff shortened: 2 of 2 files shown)


lugassawan (approved):
Clean, correct, well-tested bug fix — approving.

**What the fix does**: adds an early-return guard at `rename.go:31` (`if newBranch == wt.Branch`) that fires *after* prefix resolution (so the comparison is apples-to-apples against the fully-qualified branch name) and *before* the `git.BranchExists` check. This correctly suppresses both the misleading "branch already exists" error and the destructive `git branch -D` recovery hint that followed it.

**Why this placement is right**: resolving the prefix first means the comparison can't be confused by task-only input; short-circuiting before any git call means zero side-effects for the no-op case. Using plain `fmt.Errorf` (no `errhint.WithFix`) is correct — there's nothing to recover from.

**Test quality**: `TestRenameWorktreeSameName` is strong — it asserts the error message text, the *absence* of `git branch -D` / `To fix:` in the output (proving no recovery hint leaks through), and `calls == 0` (proving the guard short-circuits before any git invocation). All 6 rename tests pass.

One non-blocking design note: same-name rename returns an error rather than a silent no-op — defensible choice, and consistent with the intent described in #220.

Small, focused, no unrelated changes, no new dependencies. Thanks @ibobgunardi!

PR #263, opened 2026-06-05
[feat] consent gate before executing committed shell commands (#253)

lugassawan:
## Summary

- Adds a direnv-style consent gate that blocks `rimba` from executing committed shell commands (`post_create`, `post_rename`, `deps.modules[].install`) without explicit per-repo user approval
- New `internal/trust` package: content-based SHA256 hash of the command set, per-repo `.rimba/trust.local.toml` store (gitignored, 0600 perms)
- `ensureTrust()` gate inserted at every command entry point that triggers shell execution (`add`, `rename`, `deps install`, `duplicate`, `restore`); `branch:` promote mode is correctly ungated (no hooks run there)
- `rimba trust` command with `--show` flag to review commands and approval status
- Non-interactive gate (`GateNonInteractive`) for MCP tool handlers (auto-deny unless pre-trusted or `RIMBA_TRUST_YES`)
- `--yes` persistent flag and `RIMBA_TRUST_YES=1` env var for CI escape hatch
- `rimba init` registers `trust.local.toml` in `.gitignore`; `Record()` self-heals repos initialized before this feature
- Approval is automatically invalidated when the committed command set changes (different hash → gate re-arms)

## Test Plan

- [x] Unit tests pass (`make test`) — 2056 passed
- [x] E2E tests pass (`make test-e2e`) — 309 passed
- [x] Coverage ≥ 97% — 97.0%
- [x] AC1: fresh repo + `post_create` → `rimba add` without consent exits non-zero, hook does not run
- [x] AC2: after approval, changing the command set re-arms the gate (different hash)
- [x] AC3: `RIMBA_TRUST_YES=1` and `--yes` each bypass the prompt and record approval

## Issue

Closes #253

### DIFF
--- a/cmd/duplicate.go
+++ b/cmd/duplicate.go
@@ -97,6 +97,10 @@ var duplicateCmd = &cobra.Command{
 			return fmt.Errorf("worktree path already exists: %s", wtPath)
 		}
 
+		if err := ensureTrust(cmd, repoRoot, cfg); err != nil {
+			return err
+		}
+
 		dryRun, _ := cmd.Flags().GetBool(flagDryRun)
 		skipDeps, _ := cmd.Flags().GetBool(flagSkipDeps)
 		skipHooks, _ := cmd.Flags().GetBool(flagSkipHooks)
--- a/cmd/duplicate.go
+++ b/cmd/duplicate.go
@@ -97,10 +97,6 @@ var duplicateCmd = &cobra.Command{
 			return fmt.Errorf("worktree path already exists: %s", wtPath)
 		}
 
-		if err := ensureTrust(cmd, repoRoot, cfg); err != nil {
-			return err
-		}
-
 		dryRun, _ := cmd.Flags().GetBool(flagDryRun)
 		skipDeps, _ := cmd.Flags().GetBool(flagSkipDeps)
 		skipHooks, _ := cmd.Flags().GetBool(flagSkipHooks)
@@ -125,6 +121,10 @@ var duplicateCmd = &cobra.Command{
 				fmt.Fprintf(out, "[dry-run] would run post-create hooks\n")
 			}
 			return nil
+		}
+
+		if err := ensureTrust(cmd, repoRoot, cfg); err != nil {
+			return err
 		}
 
 		s := spinner.New(spinnerOpts(cmd))
--- a/internal/trust/store.go
+++ b/internal/trust/store.go
@@ -1,0 +1,119 @@
+package trust
+
+import (
+	"errors"
+	"fmt"
+	"os"
+	"path/filepath"
+	"time"
+
+	toml "github.com/pelletier/go-toml/v2"
+
+	"github.com/lugassawan/rimba/internal/config"
+	"github.com/lugassawan/rimba/internal/errhint"
+	"github.com/lugassawan/rimba/internal/fileutil"
+)
+
+const (
+	// FileName is the gitignored local trust record for a repo.
+	FileName     = "trust.local.toml"
+	storeVersion = 1
+)
+
+// Store is the content of the per-repo trust record.
+type Store struct {
+	Version    int    `toml:"version"`
+	Hash       string `toml:"hash"`
+	ApprovedAt string `toml:"approved_at"`
+}
+
+// Load reads the trust store from repoRoot. Returns (nil, nil) if absent.
+func Load(repoRoot string) (*Store, error) {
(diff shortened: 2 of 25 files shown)


cmd/duplicate.go
```diff
@@ -97,6 +97,10 @@ var duplicateCmd = &cobra.Command{
 			return fmt.Errorf("worktree path already exists: %s", wtPath)
 		}
 
+		if err := ensureTrust(cmd, repoRoot, cfg); err != nil {
```
lugassawan: **Important:** `duplicate --dry-run` triggers the trust consent prompt even though no shell commands execute in dry-run mode.

`ensureTrust` is called at line 100, before the `dryRun` check at line 115. A user running `rimba duplicate auth --dry-run` on an unconsented repo will be blocked (or prompted) for commands that won't actually run.

Suggested fix — guard the gate behind the dry-run check:
```go
dryRun, _ := cmd.Flags().GetBool(flagDryRun)
if !dryRun {
    if err := ensureTrust(cmd, repoRoot, cfg); err != nil {
        return err
    }
}
```
Or simply move `ensureTrust` to after the dry-run early-exit block at line 115.

internal/trust/store.go
```diff
@@ -0,0 +1,119 @@
...
+		errors.New("committed shell commands are not trusted for this repo"),
+		"run 'rimba trust' to approve them, or set RIMBA_TRUST_YES=1 for CI",
+	)
+}
+
+func storePath(repoRoot string) string {
+	return filepath.Join(repoRoot, config.DirName, FileName)
+}
+
+// trustYesEnv returns true when RIMBA_TRUST_YES is set to a truthy value.
```
lugassawan: **Important:** `trustYesEnv()` in `store.go` is a private duplicate of the env-check logic in `cmd/trust_gate.go:68-74` (`trustYesRequested`). Both define the truthy semantics for `RIMBA_TRUST_YES` (`ok && v != "" && v != "0"`). If the semantics change (e.g. adding `"false"` as a falsy value), one site will be missed.

Suggested fix — export the logic from the `trust` package:
```go
// TrustYesFromEnv returns true when RIMBA_TRUST_YES is set to a truthy value.
func TrustYesFromEnv() bool {
    v, ok := os.LookupEnv("RIMBA_TRUST_YES")
    return ok && v != "" && v != "0"
}
```
Then call `trust.TrustYesFromEnv()` from both `GateNonInteractive` and `trustYesRequested`, eliminating the duplication.

internal/trust/store.go
```diff
@@ -0,0 +1,119 @@
...
+
+const (
+	// FileName is the gitignored local trust record for a repo.
+	FileName     = "trust.local.toml"
+	storeVersion = 1
+)
+
+// Store is the content of the per-repo trust record.
+type Store struct {
+	Version    int    `toml:"version"`
```
lugassawan, line 25: **Important:** `Store.Version` is written on every `Record` call (`storeVersion = 1` at line 20), but `Load()` and `IsTrusted()` never inspect it. The field is currently write-only — it persists to disk but has no effect on reads.

Two options:

**(a) Remove the field** until there's a concrete migration story.

**(b) Add a forward-compat check in `Load()`** so newer binaries writing newer store versions are recognized:
```go
if s.Version > storeVersion {
    return nil, fmt.Errorf("trust store version %d is newer than supported (%d); upgrade rimba", s.Version, storeVersion)
}
```
Option (b) is preferable — it ensures a future format change doesn't silently produce wrong behavior for users on older binaries.

cmd/duplicate.go
lugassawan: Addressed in 71527016dad232168a89d73f4fd64f6ec34aa2f4: moved `ensureTrust` to after the dry-run early-return block so `rimba duplicate --dry-run` no longer triggers the consent prompt.

internal/trust/store.go
lugassawan: Addressed in 71527016dad232168a89d73f4fd64f6ec34aa2f4: renamed `trustYesEnv` → exported `TrustYesFromEnv` in `internal/trust`; `cmd/trust_gate.go`'s `trustYesRequested` now delegates to it, eliminating the duplicate semantics.

internal/trust/store.go
lugassawan: Addressed in 71527016dad232168a89d73f4fd64f6ec34aa2f4: added `if s.Version > storeVersion` guard in `Load()` with a clear upgrade hint; covered by new `TestLoadNewerVersion`.

PR #264, opened 2026-06-05
[feat] consolidate .rimba/*.local.toml gitignore entries under a single glob

lugassawan:
## Summary
- Replace per-file `.rimba/settings.local.toml` and `.rimba/trust.local.toml` gitignore entries with a single `.rimba/*.local.toml` glob, so any current or future `*.local.toml` file under `.rimba/` is ignored without further `.gitignore` churn
- Add `trust.EnsureLocalGlobIgnored` which removes pre-existing per-file entries and adds the glob (no-op for `--personal` repos where `.rimba/` is already ignored)
- Re-running `rimba init` on an existing repo now reconciles per-file entries to the consolidated glob
- `trust.Record` also self-heals on approval, migrating repos initialized before this change

## Test Plan
- [x] Coverage ≥97% (`make test-coverage` → 97.0%)
- [x] New `HasGitignoreEntry` + `EnsureLocalGlobIgnored` tests added
- [x] Re-init migration test: seeds `.gitignore` with per-file entries, asserts they're replaced by one glob line after `rimba init`

## Issue

Issue: N/A — direct feature request, no ticket

### DIFF
--- a/cmd/init.go
+++ b/cmd/init.go
@@ -287,11 +287,9 @@ func printSection(cmd *cobra.Command, title string, tier installTier, results []
 	}
 }
 
-// ensureLocalIgnore writes the gitignore entry for local config files.
-// In personal mode it ignores the whole .rimba/ dir; otherwise it consolidates
-// all *.local.toml files under a single glob via trust.EnsureLocalGlobIgnored.
-// gitignoreEntry is only used in personal mode; in non-personal mode
-// trust.EnsureLocalGlobIgnored always derives the glob from config.LocalGlob.
+// ensureLocalIgnore updates .gitignore for local config files.
+// Personal mode ignores the whole .rimba/ dir; non-personal mode writes the
+// *.local.toml glob. gitignoreEntry is only used in the personal branch.
 func ensureLocalIgnore(repoRoot, gitignoreEntry string, personal bool) (bool, error) {
 	if personal {
 		return fileutil.EnsureGitignore(repoRoot, gitignoreEntry)
@@ -299,11 +297,8 @@ func ensureLocalIgnore(repoRoot, gitignoreEntry string, personal bool) (bool, er
 	return trust.EnsureLocalGlobIgnored(repoRoot)
 }
 
-// reconcileExistingIgnore runs on re-init of an already-initialized repo.
-// It migrates per-file gitignore entries to the consolidated glob and prints
-// a message only when the glob line was newly added.
-// NOTE: switching from non-personal to personal mode (--personal on re-init) is
-// not handled here; users must manually add .rimba/ to .gitignore if desired.
+// reconcileExistingIgnore migrates per-file .gitignore entries to the glob on
+// re-init. Switching from non-personal to personal mode is not handled here.
 func reconcileExistingIgnore(cmd *cobra.Command, repoRoot string, personal bool) error {
 	if personal {
 		return nil
--- a/cmd/init.go
+++ b/cmd/init.go
@@ -290,6 +290,8 @@ func printSection(cmd *cobra.Command, title string, tier installTier, results []
 // ensureLocalIgnore writes the gitignore entry for local config files.
 // In personal mode it ignores the whole .rimba/ dir; otherwise it consolidates
 // all *.local.toml files under a single glob via trust.EnsureLocalGlobIgnored.
+// gitignoreEntry is only used in personal mode; in non-personal mode
+// trust.EnsureLocalGlobIgnored always derives the glob from config.LocalGlob.
 func ensureLocalIgnore(repoRoot, gitignoreEntry string, personal bool) (bool, error) {
 	if personal {
 		return fileutil.EnsureGitignore(repoRoot, gitignoreEntry)
@@ -300,6 +302,8 @@ func ensureLocalIgnore(repoRoot, gitignoreEntry string, personal bool) (bool, er
 // reconcileExistingIgnore runs on re-init of an already-initialized repo.
 // It migrates per-file gitignore entries to the consolidated glob and prints
 // a message only when the glob line was newly added.
+// NOTE: switching from non-personal to personal mode (--personal on re-init) is
+// not handled here; users must manually add .rimba/ to .gitignore if desired.
 func reconcileExistingIgnore(cmd *cobra.Command, repoRoot string, personal bool) error {
 	if personal {
 		return nil
--- a/cmd/init.go
+++ b/cmd/init.go
@@ -83,10 +83,10 @@ register MCP — it only updates agent files. Registration is idempotent.`,
 
 		dirPath := filepath.Join(repoRoot, config.DirName)
 		legacyPath := filepath.Join(repoRoot, config.FileName)
-		localEntry := filepath.Join(config.DirName, config.LocalFile)
+		globEntry := filepath.Join(config.DirName, config.LocalGlob)
 		dirEntry := config.DirName + "/"
 
-		gitignoreEntry := localEntry
+		gitignoreEntry := globEntry
 		if personal {
 			gitignoreEntry = dirEntry
 		}
@@ -95,6 +95,9 @@ register MCP — it only updates agent files. Registration is idempotent.`,
 		case dirExists(dirPath):
 			// .rimba/ directory already exists
 			fmt.Fprintf(cmd.OutOrStdout(), "Config %s already exists, skipping config creation\n", dirPath)
+			if err := reconcileExistingIgnore(cmd, repoRoot, personal); err != nil {
+				return err
+			}
 
 		case fileExists(legacyPath):
 			if err := runInitMigrate(cmd, repoRoot, dirPath, legacyPath, gitignoreEntry, personal); err != nil {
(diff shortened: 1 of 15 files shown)


internal/trust/store.go
```diff
@@ -56,18 +56,32 @@ func IsTrusted(repoRoot, hash string) (bool, error) {
...
+// gitignore glob: it removes any pre-existing specific entries and adds the glob.
+// No-op when the whole .rimba/ directory is already ignored (--personal repos).
+// Returns whether the glob line was newly added.
+func EnsureLocalGlobIgnored(repoRoot string) (added bool, err error) {
+	hasDir, err := fileutil.HasGitignoreEntry(repoRoot, config.DirName+"/")
+	if err != nil || hasDir {
+		return false, err
+	}
+	_, _ = fileutil.RemoveGitignoreEntry(repoRoot, filepath.Join(config.DirName, config.LocalFile))
+	_, _ = fileutil.RemoveGitignoreEntry(repoRoot, filepath.Join(config.DirName, FileName))
```
lugassawan: **High:** Silent error discard on `RemoveGitignoreEntry` can leave `.gitignore` in a partially-mutated state. If the first removal succeeds but the second fails mid-write, the file will have `.rimba/settings.local.toml` removed but `.rimba/trust.local.toml` still present — then the glob is appended, violating the invariant that no per-file entries coexist with the glob.

The failure surface is narrow (concurrent writers, FS quota), but silent discard means the user gets no indication of partial cleanup. Since the glob appended below still *covers* the per-file files even if removal fails, the gitignore is functionally correct — the only side-effect is a redundant legacy line. If that reasoning is intentional, a brief comment documents it and prevents future readers from assuming the errors were accidentally swallowed:

```go
// Removal of legacy per-file entries is best-effort: even if removal fails,
// the glob appended below still covers both files, so gitignore correctness is
// preserved — the only residue is a redundant line git will ignore via the glob.
_, _ = fileutil.RemoveGitignoreEntry(repoRoot, filepath.Join(config.DirName, config.LocalFile))
_, _ = fileutil.RemoveGitignoreEntry(repoRoot, filepath.Join(config.DirName, FileName))
```

cmd/init.go
```diff
@@ -298,6 +287,33 @@ func printSection(cmd *cobra.Command, title string, tier installTier, results []
...
 }
 
+// ensureLocalIgnore writes the gitignore entry for local config files.
+// In personal mode it ignores the whole .rimba/ dir; otherwise it consolidates
+// all *.local.toml files under a single glob via trust.EnsureLocalGlobIgnored.
+func ensureLocalIgnore(repoRoot, gitignoreEntry string, personal bool) (bool, error) {
+	if personal {
+		return fileutil.EnsureGitignore(repoRoot, gitignoreEntry)
+	}
+	return trust.EnsureLocalGlobIgnored(repoRoot)
```
lugassawan: **Medium:** The `gitignoreEntry` parameter is effectively dead in the non-personal branch. `trust.EnsureLocalGlobIgnored(repoRoot)` always uses `filepath.Join(config.DirName, config.LocalGlob)` internally — whatever is passed in is silently ignored.

Today there's no visible bug because callers always set `gitignoreEntry = globEntry` in non-personal mode (line 89), so the printed message and the written entry are consistent. But the implicit contract is fragile: a future caller could pass a different entry and receive a misleading success message while the actual write used the hardcoded glob.

Consider either removing the parameter from non-personal callers or adding a note in the function comment:

```go
// ensureLocalIgnore writes the gitignore entry for local config files.
// In personal mode it writes gitignoreEntry (i.e. .rimba/); in non-personal mode
// it always consolidates under .rimba/*.local.toml via trust.EnsureLocalGlobIgnored —
// gitignoreEntry is unused in that branch.
```

cmd/init.go
```diff
@@ -298,6 +287,33 @@ func printSection(cmd *cobra.Command, title string, tier installTier, results []
...
+}
+
+// reconcileExistingIgnore runs on re-init of an already-initialized repo.
+// It migrates per-file gitignore entries to the consolidated glob and prints
+// a message only when the glob line was newly added.
+func reconcileExistingIgnore(cmd *cobra.Command, repoRoot string, personal bool) error {
+	if personal {
+		return nil
```
lugassawan, line 303: **Medium:** When a user re-runs `rimba init --personal` on a repo previously initialized without `--personal`, this function is a no-op — it returns `nil` immediately without adding `.rimba/` to `.gitignore` or removing the now-superseded glob. The `.gitignore` ends up with the glob (`.rimba/*.local.toml`) but not the directory entry (`.rimba/`), which may surprise users expecting full personal-mode gitignore coverage of the `.rimba/` directory.

This is a UX gap rather than a data-loss bug, and the PR description does not claim to handle the non-personal → personal transition. A doc comment noting the known limitation would prevent future confusion:

```go
// reconcileExistingIgnore runs on re-init of an already-initialized repo.
// It migrates per-file gitignore entries to the consolidated glob.
// NOTE: switching from non-personal to personal mode (--personal on re-init) is
// not handled here; users must manually add .rimba/ to .gitignore if desired.
func reconcileExistingIgnore(cmd *cobra.Command, repoRoot string, personal bool) error {
```

internal/trust/store.go
```diff
@@ -56,18 +56,32 @@ func IsTrusted(repoRoot, hash string) (bool, error) {
 	return s != nil && s.Hash == hash, nil
 }
 
+// EnsureLocalGlobIgnored consolidates the personal *.local.toml overrides
+// (settings.local.toml, trust.local.toml) under a single .rimba/*.local.toml
+// gitignore glob: it removes any pre-existing specific entries and adds the glob.
+// No-op when the whole .rimba/ directory is already ignored (--personal repos).
+// Returns whether the glob line was newly added.
+func EnsureLocalGlobIgnored(repoRoot string) (added bool, err error) {
```
lugassawan: **Low (design note):** `EnsureLocalGlobIgnored` is exported from the `trust` package but its logic is entirely about `.gitignore` file mutation — it reaches back into `fileutil` for all I/O and references `config` constants. The `trust` package is currently a transit layer between `cmd/init.go` and `fileutil`.

The current placement is pragmatic: `cmd/init.go` already imports `trust`, and moving the function to `fileutil` would create an import cycle (since `trust` imports `fileutil`). A future refactor could introduce a small `internal/gitignoreutil` package that both `trust` and `cmd` import, or the function could be un-exported and duplicated at the call sites.

Not a blocker — just worth tracking so the package boundary doesn't accumulate more gitignore logic over time.

internal/trust/store.go
lugassawan: Addressed in 1f25f82: added a comment explaining that removal is best-effort — the glob appended below covers both files regardless, so gitignore correctness is preserved even on partial failure.

lugassawan: Addressed in 1f25f82: added a comment explaining that removal is best-effort and the glob appended below preserves gitignore correctness even on partial failure.

cmd/init.go
lugassawan: Addressed in 1f25f82: updated ensureLocalIgnore comment to note that gitignoreEntry is unused in the non-personal branch.

cmd/init.go
lugassawan: Addressed in 1f25f82: added a NOTE comment to reconcileExistingIgnore documenting the non-personal to personal mode transition limitation.

internal/trust/store.go
lugassawan: Addressed in 2f31bd915e200291a980f5ceb010e801585195ce: moved EnsureLocalGlobIgnored to internal/fileutil alongside the other gitignore helpers; added config.TrustFile to break the potential import cycle.

internal/fileutil/gitignore_test.go
```diff
@@ -218,6 +220,90 @@ func TestRemoveGitignoreEntryWriteError(t *testing.T) {
...
+	if strings.Contains(content, ".rimba/settings.local.toml") {
+		t.Errorf(".gitignore should not contain per-file settings entry, got:\n%s", content)
+	}
+	if strings.Contains(content, ".rimba/trust.local.toml") {
+		t.Errorf(".gitignore should not contain per-file trust entry, got:\n%s", content)
+	}
+	if !strings.Contains(content, "node_modules") {
+		t.Errorf(".gitignore should preserve other entries, got:\n%s", content)
+	}
+}
```
lugassawan, line 305: **Low** — no unit test for `EnsureLocalGlobIgnored` when no `.gitignore` file exists (fresh repo). The constituent functions (`HasGitignoreEntry`, `RemoveGitignoreEntry`, `EnsureGitignore`) each handle the absent-file case individually, so the combined path is correct by composition — but a direct test would document the fresh-repo behavior explicitly:
```go
func TestEnsureLocalGlobIgnoredNoGitignore(t *testing.T) {
    dir := t.TempDir()
    added, err := EnsureLocalGlobIgnored(dir)
    if err != nil { t.Fatalf(errUnexpected, err) }
    if !added { t.Error("expected added=true for fresh repo with no .gitignore") }
    content := readFile(t, filepath.Join(dir, gitignoreFile))
    if !strings.Contains(content, config.DirName+"/"+config.LocalGlob) {
        t.Errorf("expected glob in new .gitignore, got: %s", content)
    }
}
```

lugassawan: Addressed in a8ce43b3bd4b8d1700fda1ec3cacb52194e15d3c: added TestEnsureLocalGlobIgnoredNoGitignore for the fresh-repo path; also deduplicated trust.FileName = config.TrustFile to eliminate the silent-drift hazard noted in the informational finding.

cmd/init_test.go
```diff
@@ -368,3 +368,48 @@ func TestInitFreshGitignoreAlreadyPresent(t *testing.T) {
...
+func TestInitReInitMigratesPerFileEntries(t *testing.T) {
+	repoDir := t.TempDir()
+
+	// Pre-create .rimba/ (already initialized) with per-file gitignore entries
+	rimbaDir := filepath.Join(repoDir, config.DirName)
+	if err := os.MkdirAll(rimbaDir, 0750); err != nil {
+		t.Fatal(err)
+	}
+	perFileEntries := filepath.Join(config.DirName, config.LocalFile) + "\n" +
+		filepath.Join(config.DirName, "trust.local.toml") + "\n"
```
lugassawan: **Low** — raw string literal `"trust.local.toml"` in the fixture setup (line 381) and assertion (line 412) bypasses `config.TrustFile`, undermining the single-source-of-truth invariant this PR establishes. A future rename of the constant won't update these literals.

Replace with:
```go
// line 380-381
perFileEntries := filepath.Join(config.DirName, config.LocalFile) + "\n" +
    filepath.Join(config.DirName, config.TrustFile) + "\n"

// line 412
if strings.Contains(content, filepath.Join(config.DirName, config.TrustFile)) {
```

lugassawan: Addressed in 412bbad: replaced raw "trust.local.toml" literals in TestInitReInitMigratesPerFileEntries with config.TrustFile constant.

PR #289, opened 2026-06-05
[test] Cover DepsConcurrency

ibobgunardi:
Adds table-driven coverage for `Config.DepsConcurrency()`.

The test covers both branches from #272:
- nil `Deps` returns `0`
- configured `Deps.Concurrency` returns that value

Tests:
- `go test ./internal/config -run 'Test(IsAutoDetectDeps|DepsConcurrency)' -count=1 -v`

I also tried `go test ./internal/config -count=1` on Windows. It fails in unrelated existing tests around Unix-style read-only directories and absolute paths, so I kept this PR scoped to the requested coverage gap.

Fixes #272

### DIFF
--- a/internal/config/config_test.go
+++ b/internal/config/config_test.go
@@ -759,3 +759,31 @@ func TestIsAutoDetectDeps(t *testing.T) {
 		})
 	}
 }
+
+func TestDepsConcurrency(t *testing.T) {
+	tests := []struct {
+		name string
+		cfg  config.Config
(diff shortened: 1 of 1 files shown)


lugassawan (approved):
**Review Decision: APPROVE**

---
_Reviewed by `reviewer` ([swe-workbench](https://github.com/lugassawan/swe-workbench)). Posted 0 inline comments, deduped 0._

PR #290, opened 2026-06-06
[fix] Serialize gitignore updates

ibobgunardi:
Fixes #286.

`EnsureGitignore`, `RemoveGitignoreEntry`, and the local-glob migration now share a small lock around the `.gitignore` read/modify/write path. That keeps parallel rimba processes from both appending the same entry.

While touching that path, I also made the local glob use gitignore-style `/` separators consistently, and kept migration tolerant of the older Windows `\` form.

Checks:
- `go test ./internal/fileutil -run Gitignore|LocalGlob -count=1 -v`
- `go test ./internal/fileutil -run TestEnsureGitignoreConcurrent -count=3 -timeout=180s`
- `go test ./cmd -run InitFreshGitignoreAlreadyPresent|InitReInitMigratesPerFileEntries|InitMigrationFromLegacy|InitFresh|InitExistingDirConfig|InitPersonal -count=1 -timeout=180s`
- `go test ./internal/trust -run RecordGitignore -count=1 -timeout=180s`

### DIFF
--- a/internal/fileutil/gitignore.go
+++ b/internal/fileutil/gitignore.go
@@ -1,17 +1,59 @@
 package fileutil
 
 import (
+	"errors"
 	"os"
 	"path/filepath"
 	"strings"
+	"time"
 
 	"github.com/lugassawan/rimba/internal/config"
 )
+
+var gitignoreLockTimeout = 2 * time.Second
 
 // EnsureGitignore ensures that entry is present as a line in the .gitignore
 // file at repoRoot. If the file does not exist it is created. Returns true
 // if the entry was added, false if it was already present.
 func EnsureGitignore(repoRoot string, entry string) (added bool, retErr error) {
+	return withGitignoreLock(repoRoot, func() (bool, error) {
+		return ensureGitignoreLocked(repoRoot, entry)
+	})
+}
+
+// HasGitignoreEntry reports whether entry is present as a trimmed line in the
+// .gitignore file at repoRoot. Returns false (not error) when the file is absent.
+func HasGitignoreEntry(repoRoot, entry string) (bool, error) {
+	return hasGitignoreEntry(repoRoot, entry)
+}
+
+// EnsureLocalGlobIgnored consolidates *.local.toml overrides under a single
+// .rimba/*.local.toml gitignore glob, removing any pre-existing per-file entries.
+// No-op when .rimba/ is already ignored (--personal repos).
+// Returns whether the glob line was newly added.
+func EnsureLocalGlobIgnored(repoRoot string) (added bool, err error) {
+	return withGitignoreLock(repoRoot, func() (bool, error) {
+		hasDir, err := hasGitignoreEntry(repoRoot, config.DirName+"/")
+		if err != nil || hasDir {
+			return false, err
+		}
+		// Best-effort cleanup: the glob below covers both files even if removal fails.
+		removeGitignoreEntryVariantsLocked(repoRoot, config.DirName, config.LocalFile)
+		removeGitignoreEntryVariantsLocked(repoRoot, config.DirName, config.TrustFile)
+		return ensureGitignoreLocked(repoRoot, gitignorePattern(config.DirName, config.LocalGlob))
+	})
+}
+
+// RemoveGitignoreEntry removes entry from the .gitignore file at repoRoot.
+// Returns true if the entry was removed, false if the file doesn't exist or
+// the entry was not present.
+func RemoveGitignoreEntry(repoRoot string, entry string) (bool, error) {
+	return withGitignoreLock(repoRoot, func() (bool, error) {
+		return removeGitignoreEntryLocked(repoRoot, entry)
+	})
+}
+
+func ensureGitignoreLocked(repoRoot string, entry string) (added bool, retErr error) {
 	path := filepath.Join(repoRoot, ".gitignore")
 
 	data, err := os.ReadFile(filepath.Clean(path))
@@ -53,9 +95,7 @@ func EnsureGitignore(repoRoot string, entry string) (added bool, retErr error) {
 	return true, nil
 }
 
-// HasGitignoreEntry reports whether entry is present as a trimmed line in the
-// .gitignore file at repoRoot. Returns false (not error) when the file is absent.
-func HasGitignoreEntry(repoRoot, entry string) (bool, error) {
+func hasGitignoreEntry(repoRoot, entry string) (bool, error) {
 	path := filepath.Join(repoRoot, ".gitignore")
 	data, err := os.ReadFile(filepath.Clean(path))
 	if err != nil {
@@ -72,25 +112,7 @@ func HasGitignoreEntry(repoRoot, entry string) (bool, error) {
 	return false, nil
 }
 
-// EnsureLocalGlobIgnored consolidates *.local.toml overrides under a single
-// .rimba/*.local.toml gitignore glob, removing any pre-existing per-file entries.
-// No-op when .rimba/ is already ignored (--personal repos).
-// Returns whether the glob line was newly added.
-func EnsureLocalGlobIgnored(repoRoot string) (added bool, err error) {
-	hasDir, err := HasGitignoreEntry(repoRoot, config.DirName+"/")
-	if err != nil || hasDir {
-		return false, err
-	}
-	// Best-effort cleanup: the glob below covers both files even if removal fails.
-	_, _ = RemoveGitignoreEntry(repoRoot, filepath.Join(config.DirName, config.LocalFile))
-	_, _ = RemoveGitignoreEntry(repoRoot, filepath.Join(config.DirName, config.TrustFile))
-	return EnsureGitignore(repoRoot, filepath.Join(config.DirName, config.LocalGlob))
-}
-
-// RemoveGitignoreEntry removes entry from the .gitignore file at repoRoot.
-// Returns true if the entry was removed, false if the file doesn't exist or
-// the entry was not present.
-func RemoveGitignoreEntry(repoRoot string, entry string) (bool, error) {
+func removeGitignoreEntryLocked(repoRoot string, entry string) (bool, error) {
 	path := filepath.Join(repoRoot, ".gitignore")
 
 	data, err := os.ReadFile(filepath.Clean(path))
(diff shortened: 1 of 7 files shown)


internal/fileutil/gitignore.go
```diff
@@ -118,3 +140,47 @@ func RemoveGitignoreEntry(repoRoot string, entry string) (bool, error) {
...
+	defer func() {
+		if err := unlock(); retErr == nil && err != nil && !os.IsNotExist(err) {
+			retErr = err
+		}
+	}()
+
+	return fn()
+}
+
+func acquireGitignoreLock(lockPath string) (func() error, error) {
```
lugassawan, line 188: **High — Stale lock on process crash**

`acquireGitignoreLock` creates a directory-based lock but never provides crash recovery. If the rimba process is killed (OOM, SIGKILL, panic before the `defer` in `withGitignoreLock` runs), `.gitignore.lock` remains on disk permanently. Every subsequent `rimba` call on that repo then spins for 2 s and fails with `"timed out waiting for .gitignore lock"`. Users have no recovery path without manually running `rm -rf .gitignore.lock`.

This is a strict reliability regression: the original bug produced a cosmetic duplicate line; this introduces a failure mode where *all* gitignore operations hard-fail after a crash.

**Suggested fix:** Write the current PID into a file inside the lock dir at creation time. On each failed `os.Mkdir` retry, read that PID and check if the process is still alive (`os.FindProcess` + `signal 0`). If dead, forcibly remove the stale lock and retry immediately. Alternatively, document the recovery step and consider a `rimba init --unlock` escape hatch.

internal/fileutil/gitignore.go
```diff
@@ -118,3 +140,47 @@ func RemoveGitignoreEntry(repoRoot string, entry string) (bool, error) {
 
 	return true, os.WriteFile(path, []byte(strings.Join(filtered, "\n")), 0644) //nolint:gosec // .gitignore must be world-readable for git
 }
+
+func withGitignoreLock(repoRoot string, fn func() (bool, error)) (retAdded bool, retErr error) {
+	lockPath := filepath.Join(repoRoot, ".gitignore.lock")
```
lugassawan: **High — `.gitignore.lock` is not excluded from git tracking**

The lock directory is created at `$repoRoot/.gitignore.lock`. If a process crashes mid-operation and leaves the directory behind, `git status` will show `?? .gitignore.lock` as an untracked entry — which is exactly the cosmetic dirtiness that issue #286 set out to fix.

**Suggested fix:** Move the lock path inside `.rimba/` (e.g. `filepath.Join(repoRoot, config.DirName, ".gitignore.lock")`). The existing `.rimba/` gitignore entry already covers it, so a crash-leftover lock directory would be automatically ignored by git.

internal/fileutil/gitignore_test.go
```diff
@@ -116,6 +118,117 @@ func TestEnsureGitignoreIdempotent(t *testing.T) {
 	}
 }
 
+func TestEnsureGitignoreConcurrent(t *testing.T) {
+	dir := t.TempDir()
+
+	const workers = 8
```
lugassawan, line 127: **Medium — Package-level `var` mutated without synchronisation in concurrent test**

`gitignoreLockTimeout` is a plain `var` in the production package. `TestEnsureGitignoreConcurrent` writes it on the test goroutine while other tests in the same package (e.g. `TestEnsureGitignoreLockTimeout`) also mutate it. When run with `go test -race` or `-parallel`, the Go race detector will flag this as a data race.

**Suggested fix:** Replace `var gitignoreLockTimeout = 2 * time.Second` with a `sync/atomic.Int64` storing nanoseconds. Both production (`time.Duration(gitignoreLockTimeout.Load())`) and test code get race-free access without test-package tricks.

internal/fileutil/gitignore.go
```diff
@@ -1,17 +1,59 @@
...
 // if the entry was added, false if it was already present.
 func EnsureGitignore(repoRoot string, entry string) (added bool, retErr error) {
+	return withGitignoreLock(repoRoot, func() (bool, error) {
+		return ensureGitignoreLocked(repoRoot, entry)
+	})
+}
+
+// HasGitignoreEntry reports whether entry is present as a trimmed line in the
+// .gitignore file at repoRoot. Returns false (not error) when the file is absent.
+func HasGitignoreEntry(repoRoot, entry string) (bool, error) {
```
lugassawan: **Medium — `HasGitignoreEntry` is exported but deliberately unguarded: hidden check-then-act trap**

The public `HasGitignoreEntry` delegates straight to `hasGitignoreEntry` without acquiring the lock. Any caller that uses `HasGitignoreEntry` and then conditionally calls `EnsureGitignore` re-introduces the original TOCTOU. The contract is implicit: this function is safe only for *read-only* status checks, never in check-then-act patterns.

**Suggested fix:** Either make `HasGitignoreEntry` package-private (no external callers are visible in this diff), or add a doc comment: `// HasGitignoreEntry does not acquire the .gitignore lock. Never use it in a check-then-act pattern; use EnsureGitignore or EnsureLocalGlobIgnored for atomic read-modify-write.`

internal/fileutil/gitignore.go
```diff
@@ -118,3 +140,47 @@ func RemoveGitignoreEntry(repoRoot string, entry string) (bool, error) {
...
+			return nil, err
+		}
+		if time.Now().After(deadline) {
+			return nil, errors.New("timed out waiting for .gitignore lock")
+		}
+		time.Sleep(10 * time.Millisecond)
+	}
+}
+
+func gitignorePattern(dir, file string) string {
```
lugassawan: **Low — `gitignorePattern` helper doesn't pull its weight**

```go
func gitignorePattern(dir, file string) string {
    return dir + "/" + file
}
```

This function is called in exactly two places and its body is a single string concatenation — the same expression already appears inline everywhere else in this file. It adds a layer of naming indirection without capturing any non-obvious invariant.

**Suggested fix:** Inline `dir + "/" + file` at both call sites and delete the function.

internal/fileutil/gitignore.go
```diff
@@ -118,3 +140,47 @@ func RemoveGitignoreEntry(repoRoot string, entry string) (bool, error) {
...
+	return dir + "/" + file
+}
+
+func removeGitignoreEntryVariantsLocked(repoRoot, dir, file string) {
+	entry := gitignorePattern(dir, file)
+	_, _ = removeGitignoreEntryLocked(repoRoot, entry)
+	if legacyEntry := dir + "\\" + file; legacyEntry != entry {
```
lugassawan: **Low — Legacy backslash guard is vacuously true**

```go
if legacyEntry := dir + "\\\\"; legacyEntry != entry {
```

On every platform `dir + "/" + file != dir + "\\" + file`, so this guard is always true and the double-cleanup always runs. The intent (cross-platform migration, not a runtime OS branch) is correct but the condition is misleading.

**Suggested fix:** Remove the guard and call `removeGitignoreEntryLocked` for both forms unconditionally, or add a comment: `// backslash form is always different on any OS — intentional cross-platform migration`.

lugassawan (commented):
**Review Decision: COMMENT**

---
_Reviewed by `reviewer` ([swe-workbench](https://github.com/lugassawan/swe-workbench)). Posted 6 inline comments, deduped 0._

internal/fileutil/gitignore.go
lugassawan: Fixed in 6f1306f (hybrid PID-liveness + stale-age reclaim), then hardened in 13e71a0 after a follow-up review found the reclaim itself had no ownership fencing — a slow-but-alive holder's own `unlock()` could delete a lock that had since been reclaimed and re-acquired by a different process. Each acquisition now writes a unique `pid:nanotimestamp` token; both `unlock()` and the reclaim path re-check that token immediately before deleting the lock dir, so a stale unlock can no longer destroy someone else's active lock. See `TestStaleOwnerUnlockDoesNotDeleteReclaimedLock` for the regression test that reproduces the original race and asserts it's now closed.

internal/fileutil/gitignore.go
lugassawan: Addressed in two passes. 6f1306f initially moved the lock under `.rimba/.gitignore.lock`, but a follow-up review reproduced that this only avoided `git status` noise in `--personal` mode — in default (non-personal) mode, `.gitignore` only has the `.rimba/*.local.toml` glob, which doesn't match a subdirectory. 13e71a0 fixes this by renaming the lock dir to `gitignore-lock.local.toml` — matching that existing glob exactly, so it's covered in both modes with no new gitignore entry needed. 6b150ee adds `TestGitignoreLockDirHiddenByRealGit`, which shells out to a real git repo and asserts `git status --porcelain` is empty (stronger proof than a Go-side glob-match assertion). One narrow residual is documented on `withGitignoreLock`: a crash between creating the lock dir and the very first write of the covering glob in a given repo can still leave a brief uncovered window; it self-heals on the next successful call. A `.git/`-based lock location was considered and rejected — `.git` is a plain file, not a directory, inside every linked worktree, which is this tool's core use case.

internal/fileutil/gitignore_test.go
lugassawan: Fixed in 6f1306f: `gitignoreLockTimeout` (and the new `gitignoreStaleLockAge`) are now `atomic.Int64` storing nanoseconds. All test call sites (this test, `TestEnsureGitignoreLockTimeout`, and the new reclaim tests) use `.Load()`/`.Store()` instead of bare assignment. Verified clean under `go test ./internal/fileutil/... -race -count=5`.

internal/fileutil/gitignore.go
lugassawan: Fixed in 6f1306f: confirmed via repo-wide grep that `HasGitignoreEntry` had zero callers outside `internal/fileutil` (only its own tests), so it's now unexported to `hasGitignoreEntry` with a doc comment warning that pairing it with a later Ensure/Remove call re-opens the TOCTOU window.

internal/fileutil/gitignore.go
lugassawan: Fixed in 6f1306f: inlined `dir + "/" + file` at both call sites and deleted the helper.

internal/fileutil/gitignore.go
lugassawan: Fixed in 6f1306f: removed the vacuous guard, `removeGitignoreEntryVariantsLocked` now calls `removeGitignoreEntryLocked` unconditionally for both the "/" and legacy "\\" forms, with a comment explaining why the backslash form is still checked at all.

PR #291, opened 2026-06-06
[fix] Avoid ambiguous bare task resolution

ibobgunardi:
## Summary
- check task-normalized branch matches before the exact bare-branch fallback
- return no resolver match when a bare branch collides with service-scoped task matches
- add a regression test for the collision case

Fixes #266

## Tests
- go test ./internal/resolver -run TestFindBranchForTask|TestFindAllBranchesForTask -count=1

Note: `go test ./internal/resolver` currently fails on Windows in existing `TestWorktreePath`, which expects POSIX separators.

### DIFF
--- a/internal/resolver/resolver.go
+++ b/internal/resolver/resolver.go
@@ -116,15 +116,18 @@ func FindBranchForTask(service, task string, worktrees []WorktreeInfo, prefixes 
 		return wt, true
 	}
 
-	if wt, ok := findExact(task, worktrees); ok {
-		return wt, true
-	}
-
 	if service == "" {
 		matches := FindAllBranchesForTask(task, worktrees, prefixes)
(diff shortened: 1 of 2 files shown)


lugassawan (approved):
**Review Decision: APPROVE**

---
_Reviewed by `reviewer` ([swe-workbench](https://github.com/lugassawan/swe-workbench)). Posted 0 inline comments, deduped 0._

Medium | `internal/resolver/resolver.go:129` | `findExact` reachable for `service != ""` calls without ambiguity guard | When `service` is non-empty, the `if service == ""` block is skipped entirely, so `findExact` can silently resolve a bare branch that collides with the intended `service/prefix/task` branch. Pre-existing behavior outside this PR's scope — no action needed now.

PR #293, opened 2026-06-07
[test] Cover presentation helper output

ibobgunardi:
Closes #273.

Adds direct output tests for:
- `printWorktreeResult` copied, missing, and symlink-skipped rows
- `printSyncDryRun` merge dry-run output

Checks:
- `gofmt -w cmd/add_test.go cmd/sync_test.go`
- `git diff --check`

Note: I could not run `go test` locally because this host has Go 1.25.1 while `go.mod` requires Go 1.26.4, and automatic toolchain switching hung in the sandbox.

### DIFF
--- a/cmd/add_test.go
+++ b/cmd/add_test.go
@@ -9,6 +9,7 @@ import (
 	"testing"
 
 	"github.com/lugassawan/rimba/internal/config"
+	"github.com/lugassawan/rimba/internal/operations"
 	"github.com/lugassawan/rimba/testutil"
 	"github.com/spf13/cobra"
 )
@@ -53,6 +54,32 @@ func makeWorktreeGitRunner(repoDir string) *mockRunner {
 			return "", nil
 		},
 		runInDir: noopRunInDir,
+	}
+}
+
+func TestPrintWorktreeResultCopySkipBranches(t *testing.T) {
+	cmd, buf := newTestCmd()
+
+	printWorktreeResult(cmd, "Created worktree", operations.AddResult{
+		Branch:          "feature/presentation",
+		Path:            "/repo/.rimba/worktrees/presentation",
+		Copied:          []string{".env", ".tool-versions"},
+		Skipped:         []string{".missing"},
+		SkippedSymlinks: []string{".config/link"},
+	})
+
+	got := buf.String()
+	for _, want := range []string{
+		"Created worktree\n",
+		"  Branch: feature/presentation\n",
+		"  Path:   /repo/.rimba/worktrees/presentation\n",
+		"  Copied: [.env .tool-versions]\n",
+		"  Skipped (not found): [.missing]\n",
+		"  Skipped (symlinks): [.config/link]\n",
+	} {
+		if !strings.Contains(got, want) {
+			t.Fatalf("output missing %q:\n%s", want, got)
+		}
 	}
 }
 
--- a/cmd/sync_test.go
+++ b/cmd/sync_test.go
@@ -38,6 +38,18 @@ func testSyncWorktrees() []resolver.WorktreeInfo {
 	}
 }
 
+func TestPrintSyncDryRunMerge(t *testing.T) {
+	cmd, buf := newTestCmd()
+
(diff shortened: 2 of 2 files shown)


cmd/add_test.go
```diff
@@ -56,6 +57,32 @@ func makeWorktreeGitRunner(repoDir string) *mockRunner {
 	}
 }
 
+func TestPrintWorktreeResultCopySkipBranches(t *testing.T) {
+	cmd, buf := newTestCmd()
+
```
lugassawan, line 62: **Medium** — Test name `TestPrintWorktreeResultCopySkipBranches` is misleading.

The suffix `Branches` does not describe what is tested. The test covers `Copied`, `Skipped`, and `SkippedSymlinks` output lines — no branch-plural concept is involved. A reader scanning test names builds a wrong mental model.

**Suggested fix:** Rename to `TestPrintWorktreeResultCopyAndSkipped` or `TestPrintWorktreeResultAllFields`.

cmd/sync_test.go
```diff
@@ -38,6 +38,18 @@ func testSyncWorktrees() []resolver.WorktreeInfo {
...
+	cmd, buf := newTestCmd()
+
+	printSyncDryRun(cmd, "feature/login", "main", true, true)
+
+	want := "[dry-run] would merge feature/login onto main\n" +
+		"[dry-run] would push feature/login to origin\n"
+	if got := buf.String(); got != want {
+		t.Fatalf("output = %q, want %q", got, want)
+	}
+}
```
lugassawan, line 51: **Medium** — The `push=false` branch of `printSyncDryRun` is not tested.

The guard `if push { ... }` (in `cmd/sync.go`) is the only conditional branch in the function, yet this test covers only `push=true`. A regression that unconditionally prints the push line would not be caught.

**Suggested fix:** Add a second test case, e.g.:

```go
func TestPrintSyncDryRunRebaseNoPush(t *testing.T) {
    cmd, buf := newTestCmd()
    printSyncDryRun(cmd, "feature/login", "main", false, false)
    got := buf.String()
    if !strings.Contains(got, "would rebase") {
        t.Fatalf("expected rebase verb, got: %q", got)
    }
    if strings.Contains(got, "would push") {
        t.Fatalf("push line must be absent, got: %q", got)
    }
}
```

cmd/add_test.go
```diff
@@ -56,6 +57,32 @@ func makeWorktreeGitRunner(repoDir string) *mockRunner {
...
+	printWorktreeResult(cmd, "Created worktree", operations.AddResult{
```
lugassawan, line 63: **Low** — The empty-fields path of `printWorktreeResult` is not tested.

The three `len > 0` guards (for `Copied`, `Skipped`, `SkippedSymlinks`) are untested with the current test. A complementary test would verify those lines are correctly suppressed when the slices are nil.

**Suggested fix:** Add `TestPrintWorktreeResultMinimal` with an `AddResult` where all slice fields are nil, and assert the output contains only the header, Branch, and Path lines (and does NOT contain "Copied:", "Skipped", or "symlinks").

lugassawan (dismissed):

lugassawan (commented):
**Review Decision: COMMENT**

---
_Reviewed by `reviewer` ([swe-workbench](https://github.com/lugassawan/swe-workbench)). Posted 3 inline comments, deduped 0._

cmd/add_test.go
lugassawan: Renamed to `TestPrintWorktreeResultAllFields` in f4fb3ba — thanks for catching the misleading name.

cmd/sync_test.go
lugassawan: Added `TestPrintSyncDryRunRebaseNoPush` in f4fb3ba, matching your suggested implementation exactly.

cmd/add_test.go
lugassawan: Added `TestPrintWorktreeResultMinimal` in f4fb3ba covering the nil-slice path for Copied/Skipped/SkippedSymlinks.

PR #296, opened 2026-06-10
[fix] Pin updater download URLs to trusted GitHub hosts (#281)

lugassawan:
## Summary
- Add `allowedAssetHosts` allowlist (`github.com` + `objects.githubusercontent.com`) and `validateAssetURL` helper in `internal/updater/updater.go`
- Call `validateAssetURL` in `Check()` after the empty-URL guards, before assigning `DownloadURL`/`ChecksumsURL` — a MITM or DNS-hijack that tampers with the GitHub releases API response can no longer redirect the binary or `checksums.txt` to an attacker-controlled host
- Uses `url.Hostname()` for exact host matching (rejects `github.com.evil.com`, strips ports)
- Fails closed: returns an error immediately, no download occurs
- Replace the stale comment in `findReleaseAssets` that incorrectly argued checksum verification alone was sufficient mitigation

## Test Plan

New tests:
- `TestValidateAssetURL` (7 cases: both happy-path hosts, http scheme, untrusted host, subdomain spoof, unparseable URL, empty string)
- `TestCheckRejectsUntrustedDownloadHost` — evil.com download URL → Check errors "untrusted download URL"
- `TestCheckRejectsUntrustedChecksumsHost` — evil.com checksums URL → Check errors "untrusted checksums URL"

All `example.com` test fixtures migrated to `https://github.com/...` URLs so they pass through the new validation gate.

## Issue
Closes #281

### DIFF
--- a/internal/updater/updater.go
+++ b/internal/updater/updater.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"net/url"
 	"os"
 	"path/filepath"
 	"runtime"
@@ -25,6 +26,16 @@ const (
 	checksumsFileName  = "checksums.txt"
 	goosWindows        = "windows"
 )
+
+// allowedAssetHosts pins release-asset downloads to GitHub-controlled hosts.
+// The GitHub API's browser_download_url is always https://github.com/...; the
+// CDN redirect to objects.githubusercontent.com is followed at fetch time and
+// never appears in the API response. objects.githubusercontent.com is included
+// defensively for any future direct CDN URLs that GitHub may introduce.
+var allowedAssetHosts = map[string]bool{
+	"github.com":                    true,
+	"objects.githubusercontent.com": true,
+}
 
 // maxBinarySize is the decompressed size limit for the extracted binary (100 MiB).
 // Tests may override this to exercise the size-limit path.
@@ -137,6 +148,12 @@ func (u *Updater) Check(ctx context.Context) (*CheckResult, error) {
 		}
 		if checksumsURL == "" {
 			return nil, fmt.Errorf("%s not found in release %s", checksumsFileName, release.TagName)
+		}
+		if err := validateAssetURL(downloadURL); err != nil {
+			return nil, fmt.Errorf("untrusted download URL: %w", err)
+		}
+		if err := validateAssetURL(checksumsURL); err != nil {
+			return nil, fmt.Errorf("untrusted checksums URL: %w", err)
 		}
 		result.DownloadURL = downloadURL
 		result.AssetName = assetName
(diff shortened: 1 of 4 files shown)


internal/updater/updater.go
```diff
@@ -368,3 +383,23 @@ func findReleaseAssets(assetName string, assets []Asset) (downloadURL, checksums
 	}
 	return
 }
+
+// validateAssetURL rejects asset URLs that are not https or whose host is not a
+// trusted GitHub host, defending against a tampered API response that redirects
+// the binary and checksums.txt to an attacker-controlled server (issue #281).
+func validateAssetURL(rawURL string) error {
+	if rawURL == "" {
```
lugassawan, line 387: The empty-string guard is defensively sound — but since `Check()` already pre-screens both URLs at lines 146–151, this path is unreachable in the production call path. Worth a one-line comment so future callers understand the contract:

```go
// Defensive: Check() pre-screens for empty, but callers outside Check may not.
if rawURL == "" {
```

Alternatively, if the intent is that `validateAssetURL` is always guarded by the call site, drop the check here and document the precondition in the function doc.

internal/updater/updater_test.go
```diff
@@ -1534,3 +1534,97 @@ func TestAssetNameFor(t *testing.T) {
...
+			url:     "http://github.com/lugassawan/rimba/releases/download/v2.0.0/rimba_2.0.0_linux_amd64.tar.gz",
+			wantErr: "must use https",
+		},
+		{
+			name:    "untrusted host rejected",
+			url:     "https://evil.com/lugassawan/rimba/releases/download/v2.0.0/rimba_2.0.0_linux_amd64.tar.gz",
+			wantErr: "not a trusted GitHub host",
+		},
+		{
+			name:    "subdomain spoof rejected",
```
lugassawan, line 1565: Since `objects.githubusercontent.com` is in the allowlist, it's worth adding a parallel spoof test to document that suffix-only matching doesn't apply to it either:

```go
{
    name:    "objects.githubusercontent.com subdomain spoof rejected",
    url:     "https://evil.objects.githubusercontent.com/rimba_2.0.0_linux_amd64.tar.gz",
    wantErr: "not a trusted GitHub host",
},
```

Not a blocker — the exact-match map already rejects this correctly — but it closes the documentation gap.

internal/updater/updater.go
lugassawan: Addressed in c4b39a0b584a0d10ba80edb9d5d3dde25af0ad39: added `// Defensive: Check() pre-screens for empty, but callers outside Check may not.` above the guard.

internal/updater/updater_test.go
lugassawan: Addressed in c4b39a0b584a0d10ba80edb9d5d3dde25af0ad39: added `objects.githubusercontent.com subdomain spoof rejected` test case (`evil.objects.githubusercontent.com`).

PR #298, opened 2026-06-10
[fix] Require config for MCP cleanup tools

Jia-Ethan:
## Summary
- Require initialized rimba config before MCP `clean`, `remove`, and `status` handlers do any work.
- Reuse the loaded config's default source for `clean`/`status` main-branch resolution.
- Add regression coverage for the uninitialized-repo path on all three MCP tools.

## Test Plan

Additional focused checks run:

```bash
GOPROXY=https://goproxy.cn,direct go test ./internal/mcp -run 'RequiresConfig' -count=1
GOPROXY=https://goproxy.cn,direct go test ./internal/mcp -count=1
GOPROXY=https://goproxy.cn,direct go test ./... -count=1
```

Risk and scope: scoped to MCP handler preflight checks and tests. Existing required-argument errors still run before the config gate, while valid tool calls now fail with the same friendly initialization message used by the other MCP tools when `.rimba/settings.toml` is unavailable.

## Issue

Closes #267

### DIFF
--- a/internal/mcp/tool_clean.go
+++ b/internal/mcp/tool_clean.go
@@ -38,6 +38,11 @@ func handleClean(hctx *HandlerContext) server.ToolHandlerFunc {
 			return mcp.NewToolResultError("mode is required"), nil
 		}
 
+		cfg, cfgErr := hctx.requireConfig()
+		if cfgErr != nil {
+			return mcp.NewToolResultError(cfgErr.Error()), nil
+		}
+
 		dryRun := req.GetBool("dry_run", false)
 		force := req.GetBool("force", false)
 		staleDays := req.GetInt("stale_days", 14)
@@ -48,9 +53,9 @@ func handleClean(hctx *HandlerContext) server.ToolHandlerFunc {
 		case "prune":
 			return mcpCleanPrune(ctx, r, dryRun)
 		case "merged":
-			return mcpCleanMerged(ctx, r, hctx, dryRun, force)
+			return mcpCleanMerged(ctx, r, cfg.DefaultSource, dryRun, force)
 		case "stale":
-			return mcpCleanStale(ctx, r, hctx, dryRun, staleDays, force)
+			return mcpCleanStale(ctx, r, cfg.DefaultSource, dryRun, staleDays, force)
 		default:
 			return mcp.NewToolResultError(fmt.Sprintf("invalid mode %q; use prune, merged, or stale", mode)), nil
 		}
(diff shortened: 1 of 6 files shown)


internal/mcp/tool_clean.go
```diff
@@ -38,6 +38,11 @@ func handleClean(hctx *HandlerContext) server.ToolHandlerFunc {
 			return mcp.NewToolResultError("mode is required"), nil
 		}
 
+		cfg, cfgErr := hctx.requireConfig()
```
lugassawan, line 41: **High — `prune` mode now unnecessarily requires config**

`mcpCleanPrune` calls only `git.Prune` and `git.PruneRemotes` — it never reads `cfg.DefaultSource`. By placing `requireConfig()` before the `switch`, all three modes (including `prune`) now fail without an initialized repo, extending the blast radius beyond the issue scope.

`prune` was previously usable in any repo; this is a silent behavioral regression.

**Suggested fix:** move the config gate inside the `merged` and `stale` case arms, where `cfg.DefaultSource` is actually consumed:

```go
switch mode {
case "prune":
    return mcpCleanPrune(ctx, r, dryRun)
case "merged":
    cfg, cfgErr := hctx.requireConfig()
    if cfgErr != nil {
        return mcp.NewToolResultError(cfgErr.Error()), nil
    }
    return mcpCleanMerged(ctx, r, cfg.DefaultSource, dryRun, force)
case "stale":
    cfg, cfgErr := hctx.requireConfig()
    if cfgErr != nil {
        return mcp.NewToolResultError(cfgErr.Error()), nil
    }
    return mcpCleanStale(ctx, r, cfg.DefaultSource, dryRun, staleDays, force)
```

Then update `TestCleanToolRequiresConfig` to use `modeMerged` or `modeStale` (since `modePrune` would no longer require config).

lugassawan (commented):
**Review Decision: COMMENT**

_Reviewed by `reviewer` ([swe-workbench](https://github.com/lugassawan/swe-workbench)). Posted 1 inline comment, deduped 0._

**`internal/mcp/tool_list.go` — `handleListArchived` bypasses the config gate (pre-existing)**

`handleList` short-circuits to `handleListArchived` before calling `requireConfig()` when `archived=true`. `handleListArchived` uses the nil-safe `configDefault(hctx)` helper, so `list --archived` still works without an initialized repo. This PR does not introduce or worsen it — it predates this change — but it is the same class of gap issue #267 addressed on the other three tools. Worth a follow-up issue.

PR #299, opened 2026-06-10
[fix] Thread context through the git package (#283)

lugassawan:
## Summary
- Thread ctx context.Context (first param) through all git package functions, replacing r.Run/r.RunInDir calls with r.Run(ctx, ...)/r.RunInDir(ctx, ...), so SIGINT and MCP per-request deadlines cancel in-flight git subprocesses.
- Remove Run/RunInDir (the context.Background() shims) from the Runner interface, ExecRunner, and the debug decorator.
- Rename RunContext->Run and RunInDirContext->RunInDir: since the shims are gone, the Context suffix is no longer needed.
- Recovery/rollback paths (AbortRebase, MergeAbort, StashApply/StashDrop, stashRefBySHA, MoveWorktree) deliberately stay non-cancellable via r.RunInDir(context.Background(), ...), each with a doc comment.

## Test Plan
- [x] go build clean
- [x] go fmt clean
- [x] make lint 0 issues
- [x] go test 2091 tests pass in 27 packages
- [x] make test-race passes
- [x] sync_ctx_test.go verifies cancellation propagation

Closes #283

### DIFF
--- a/cmd/init.go
+++ b/cmd/init.go
@@ -1,6 +1,7 @@
 package cmd
 
 import (
+	"context"
 	"fmt"
 	"os"
 	"path/filepath"
@@ -73,7 +74,8 @@ register MCP — it only updates agent files. Registration is idempotent.`,
 
 		r := newRunner()
 
-		repoRoot, err := git.RepoRoot(r)
+		ctx := cmd.Context()
+		repoRoot, err := git.RepoRoot(ctx, r)
 		if err != nil {
 			return err
 		}
@@ -104,7 +106,7 @@ register MCP — it only updates agent files. Registration is idempotent.`,
 			}
 
 		default:
-			if err := runInitFresh(cmd, r, repoRoot, dirPath, gitignoreEntry, personal); err != nil {
+			if err := runInitFresh(cmd, ctx, r, repoRoot, dirPath, gitignoreEntry, personal); err != nil {
 				return err
 			}
 		}
@@ -118,13 +120,13 @@ register MCP — it only updates agent files. Registration is idempotent.`,
 	},
 }
 
-func runInitFresh(cmd *cobra.Command, r git.Runner, repoRoot, dirPath, gitignoreEntry string, personal bool) error {
-	repoName, err := git.RepoName(r)
+func runInitFresh(cmd *cobra.Command, ctx context.Context, r git.Runner, repoRoot, dirPath, gitignoreEntry string, personal bool) error {
+	repoName, err := git.RepoName(ctx, r)
 	if err != nil {
 		return err
 	}
 
-	defaultBranch, err := git.DefaultBranch(r)
+	defaultBranch, err := git.DefaultBranch(ctx, r)
 	if err != nil {
 		return err
 	}
--- a/cmd/init.go
+++ b/cmd/init.go
@@ -106,7 +106,7 @@ register MCP — it only updates agent files. Registration is idempotent.`,
 			}
 
 		default:
-			if err := runInitFresh(cmd, ctx, r, repoRoot, dirPath, gitignoreEntry, personal); err != nil {
+			if err := runInitFresh(ctx, cmd, r, repoRoot, dirPath, gitignoreEntry, personal); err != nil {
 				return err
 			}
 		}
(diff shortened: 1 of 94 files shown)


internal/operations/add_branch.go
```diff
@@ -109,7 +109,9 @@ func validateForPromotion(r git.Runner, repoRoot, branch string) (string, error)
 
 // restoreStash re-applies and drops a stash entry to undo a failed mid-operation stash push.
 // No-ops when sha is empty. Returns an error if the apply fails (stash entry is preserved).
-func restoreStash(r git.Runner, dir, sha string) error {
+// StashApply and StashDrop are intentionally non-cancellable (recovery paths must complete).
+func restoreStash(ctx context.Context, r git.Runner, dir, sha string) error {
```
lugassawan: **Medium** — `_ = ctx` is misleading here.

`restoreStash` and `applyStashToWorktree` accept `ctx context.Context` but immediately discard it with `_ = ctx`. The established pattern in this PR for non-cancellable recovery paths is to drop the ctx parameter entirely (see `StashApply`, `StashDrop`, `MoveWorktree`, `AbortRebase`, `MergeAbort`). Accepting ctx implies it flows through — it doesn't, and that could mislead a future maintainer into thinking cancellation is respected here.

**Suggested fix:** Remove `ctx context.Context` from both function signatures and drop the `_ = ctx` lines. Call sites in `PromoteBranch` just don't pass ctx, mirroring how `StashApply`/`StashDrop` are called.

internal/debug/debug_test.go
```diff
@@ -9,28 +9,16 @@ import (
 )
 
 type stubRunner struct {
-	runCalled         bool
-	runInDirCalled    bool
 	runContextCalled  bool
```
lugassawan: **Low** — Field names `runContextCalled` / `runInDirCtxCalled` are stale after the rename.

`stubRunner` tracks whether `Run` was called via `runContextCalled` and whether `RunInDir` was called via `runInDirCtxCalled` (lines 12–13, 17, 22, 104, 122). The `Context` suffix no longer matches the method names, which were renamed to `Run`/`RunInDir` in this PR.

**Suggested fix:** Rename the fields to `runCalled` / `runInDirCalled` throughout `debug_test.go` for symmetry with the method names.

internal/operations/add_branch.go
lugassawan: Addressed in 77b7960626bef414a3d2dcfbd7b2b61eff27e06f: removed `ctx context.Context` param and `_ = ctx` discard from `restoreStash` and `applyStashToWorktree` — these recovery functions now match the no-ctx pattern of `StashApply`/`StashDrop`/`MoveWorktree`.

internal/debug/debug_test.go
lugassawan: Addressed in 77b7960626bef414a3d2dcfbd7b2b61eff27e06f: renamed `runContextCalled`→`runCalled` and `runInDirCtxCalled`→`runInDirCalled` throughout `debug_test.go`.

cmd/init.go
```diff
@@ -118,13 +120,13 @@ register MCP — it only updates agent files. Registration is idempotent.`,
 	},
 }
 
-func runInitFresh(cmd *cobra.Command, r git.Runner, repoRoot, dirPath, gitignoreEntry string, personal bool) error {
-	repoName, err := git.RepoName(r)
+func runInitFresh(cmd *cobra.Command, ctx context.Context, r git.Runner, repoRoot, dirPath, gitignoreEntry string, personal bool) error {
```
lugassawan: **Low** — `ctx context.Context` placed after `cmd *cobra.Command` in `runInitFresh` signature.

Go convention (and the style used everywhere else in this PR) puts `ctx context.Context` as the first parameter. This signature has it second, after `cmd`:

```go
func runInitFresh(cmd *cobra.Command, ctx context.Context, r git.Runner, ...) error
```

**Suggested fix:** Reorder to put `ctx` first:
```go
func runInitFresh(ctx context.Context, cmd *cobra.Command, r git.Runner, ...) error
```

Update the call site at line 109 accordingly: `runInitFresh(ctx, cmd, r, ...)`.

This does not affect runtime behavior — it's a style inconsistency only.

lugassawan: Addressed in 0c7d3b772d6ea4b3f1963ebdac2394050798047b: reordered `runInitFresh` signature to `(ctx context.Context, cmd *cobra.Command, r git.Runner, ...)` — ctx is now first, consistent with Go convention and the rest of the PR.

PR #301, opened 2026-06-10
[fix] Close goroutine leak in update hint + add exec cobra-level test seam

lugassawan:
## Summary

- **Finding 1 — goroutine/socket leak:** `checkUpdateHint` called `u.Check(context.Background())`, so the inner HTTP goroutine kept running for up to the updater's 30 s client timeout after the outer 2 s window elapsed. Added `ctx context.Context` param, derived `context.WithTimeout(ctx, timeout)` for the request, and passed `tctx` to `u.Check` so the HTTP round-trip is bounded by both the caller's context and the timeout. Also reset `rootCmd.ctx` in `TestExecute`'s cleanup to prevent the post-`Execute` cancelled signal-context from leaking into subsequent tests.
- **Finding 2 — missing cobra-level test seam:** Extracted `runExec(cmd, args, r git.Runner, execFn execRunner) error` from `execCmd.RunE` and added `buildExecCmd(r, execFn)` (mirrors `buildTrustCmd`, `trust.go:49`). Both production `RunE` and the test constructor delegate to `runExec`, so tests exercise the real flag-parse → filter → executor wiring. Added four cobra-level tests: `TestExecCmdMissingTargetFlag`, `TestExecCmdInvokesExecutor`, `TestExecCmdJSONOutput`, `TestExecCmdNoMatchingWorktrees`.

## Test Plan

- [x] Race detector clean (`go test -race ./...`)
- [x] Coverage ≥ 97% (`make test-coverage` → 97.2%)
- [x] New `TestCheckUpdateHintCancelledContext` verifies pre-cancelled context returns quickly without calling the HTTP server
- [x] New `TestExecCmd*` tests exercise the cobra-level wiring end-to-end with injected mocks

## Issue

Closes #287

### DIFF
--- a/cmd/exec.go
+++ b/cmd/exec.go
@@ -33,6 +33,9 @@ const (
 	hintConcurrency = "Limit the number of parallel executions"
 )
 
+// execRunner is the injectable executor function type, matching executor.Run.
+type execRunner func(context.Context, executor.Config) []executor.Result
+
 var execCmd = &cobra.Command{
 	Use:   "exec <command>",
 	Short: "Run a shell command across worktrees",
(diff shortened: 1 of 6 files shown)


cmd/update_hint.go
```diff
@@ -14,30 +14,36 @@ import (
...
 			close(ch)
 			return
 		}
 		ch <- result
 	}()
 
-	// Return a channel that resolves with the result or nil after timeout
+	// Return a channel that resolves with the result or nil when tctx expires.
 	out := make(chan *updater.CheckResult, 1)
 	go func() {
```
lugassawan, line 48: **Medium** — `cancel` is only deferred in the inner goroutine (goroutine 1); the relay goroutine (goroutine 2) exits via `tctx.Done()` without guaranteeing `cancel` is called.

`context.WithTimeout` documentation says the cancel function should be called as soon as operations running in this Context complete. When goroutine 2 exits via the `tctx.Done()` branch, goroutine 1 is still running and hasn't called `defer cancel()` yet. While goroutine 1 will eventually call it, the internal timer resource lingers until it does. Adding `defer cancel()` here ensures the resource is released promptly on both exit paths:

```go
out := make(chan *updater.CheckResult, 1)
go func() {
    defer cancel() // ensure cancel is always called on all exit paths
    select {
    case r, ok := <-ch:
        if ok {
            out <- r
        } else {
            close(out)
        }
    case <-tctx.Done():
        close(out)
    }
}()
```

cmd/exec.go
```diff
@@ -41,48 +44,73 @@ var execCmd = &cobra.Command{
...
+
+	if isJSON(cmd) {
+		return execRenderJSON(cmd, args[0], results)
+	}
+	return execRenderText(cmd, results, prefixes)
+}
+
+// buildExecCmd constructs a testable exec command with injected deps.
+// Used by exec_test.go to exercise the full flag-parse → filter → executor pipeline.
+func buildExecCmd(r git.Runner, execFn execRunner) *cobra.Command {
```
lugassawan, line 107: **Medium** — `buildExecCmd` is a second flag-registration site that can silently diverge from `execCmd.init()`.

If a future contributor adds a flag to `execCmd` (inside `init()`) but forgets to mirror it in `buildExecCmd`, the cobra-level tests will silently test a different flag surface than production — undermining the audit's intent. The production command also registers `--type` completion (`RegisterFlagCompletionFunc`) which `buildExecCmd` omits entirely.

Extracting a shared `addExecFlags(c *cobra.Command)` helper makes both sites converge on one source of truth:

```go
func addExecFlags(c *cobra.Command) {
    c.Flags().Bool(flagAll, false, "run in all eligible worktrees")
    c.Flags().String(flagType, "", "filter by prefix type")
    c.Flags().Bool(flagDirty, false, "run only in dirty worktrees")
    c.Flags().Bool(flagFailFast, false, "stop after the first failure")
    c.Flags().Int(flagConcurrency, 0, "max parallel executions (0 = unlimited)")
    c.Flags().Bool(flagJSON, false, "")
    c.Flags().Bool(flagNoColor, false, "")
}

func buildExecCmd(r git.Runner, execFn execRunner) *cobra.Command {
    c := &cobra.Command{...}
    addExecFlags(c)
    return c
}

func init() {
    addExecFlags(execCmd)
    _ = execCmd.RegisterFlagCompletionFunc(flagType, typeFilterCompletion())
    rootCmd.AddCommand(execCmd)
}
```

cmd/root.go
```diff
@@ -110,7 +110,13 @@ func init() {
 	rootCmd.SetHelpFunc(func(cmd *cobra.Command, args []string) {
 		var hint <-chan *updater.CheckResult
 		if cmd == rootCmd {
-			hint = checkUpdateHint(version, 2*time.Second)
+			ctx := cmd.Context()
```
lugassawan, line 113: **Low** — The nil-context guard is a symptom fix coupled to the `TestExecute` cleanup added in `root_test.go`.

The guard (`if ctx == nil { ctx = context.Background() }`) handles the case where `HelpFunc` is called outside `ExecuteContext` in tests. The complementary `rootCmd.SetContext(context.Background())` cleanup in `TestExecute` handles the case where a prior test left `rootCmd.ctx` as a cancelled signal context. Both pieces need to stay — removing either would cause `TestRootHelpFunctionRoot` to fail intermittently depending on test ordering. This is just an observation that the two changes are load-bearing together.

cmd/update_hint_test.go
```diff
@@ -150,3 +151,28 @@ func TestCollectHintValue(t *testing.T) {
 		t.Errorf("LatestVersion = %q, want %q", result.LatestVersion, testVersionOther)
 	}
 }
+
+func TestCheckUpdateHintCancelledContext(t *testing.T) {
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		// A pre-cancelled context should cause the HTTP client to fail before
+		// sending a request, so the handler must never complete a response.
+		t.Error("unexpected HTTP round-trip for pre-cancelled context")
```
lugassawan: **Low** — Consider `t.Fatal` instead of `t.Error` for the "should never happen" assertion in the HTTP handler.

`t.Error` marks the test failed but lets it continue, so if the HTTP handler fires unexpectedly, the timing assertions below will still execute against potentially corrupted state. `t.Fatal` makes the intent clearer — a round-trip here is always a test logic error, not a timing concern:

```go
t.Fatal("unexpected HTTP round-trip for pre-cancelled context")
```

cmd/update_hint.go
lugassawan: Addressed in 8e0064c: added `defer cancel()` to the relay goroutine so the context timer is released promptly on all exit paths, not just the happy path.

cmd/exec.go
lugassawan: Addressed in 8e0064c: extracted `addExecFlags(c *cobra.Command)` helper shared by both `init()` and `buildExecCmd()`. Both registration sites now delegate to the same function, so a new flag added to either can't silently diverge.

cmd/root.go
lugassawan: Acknowledged — both the nil guard and the `TestExecute` cleanup are intentional and load-bearing together. Neither should be removed independently.

cmd/update_hint_test.go
lugassawan: Addressed in 8e0064c: changed `t.Error` to `t.Fatal` — a round-trip here is always a logic error, so the test should stop immediately rather than continue collecting timing data.

PR #302, opened 2026-06-10
[fix] Bound git status queries with a per-item timeout (#288)

lugassawan:
## Summary

- Extend `parallel.Collect` to accept and thread `context.Context` through to each closure — callers can now enforce per-item deadlines without captured variables
- Add `git.WithItemTimeout(ctx)` helper: 10s const, mirrors the existing `prQueryTimeout` pattern
- Apply `WithItemTimeout` to the 5 git-query fan-out sites (`status_dashboard`, `mcp/tool_status`, `cmd/log`, `cmd/exec`, `list_worktrees`); the 2 long-running sites (`deps/manager`, `mcp/tool_sync`) explicitly pass ctx through without a cap
- Add `Unknown bool` to `WorktreeStatus`: a timed-out or errored worktree now renders as gray `"unknown"` instead of silently appearing clean (`✓`)
- Make `fsutil.DirSize` cancellable: runs `filepath.WalkDir` in a goroutine+select so a stuck `stat()` syscall (stalled NFS) can't block past the deadline
- Fix `AheadBehind` to propagate `ctx.Err()` while still swallowing "no upstream" errors — previously all errors were silently treated as (0,0,nil)

## Test Plan

- [x] Unit tests pass (`make test`) — 2104 tests, all pass
- [x] Coverage ≥ 97% (`make test-coverage`) — 97.3%
- [x] `TestWithItemTimeout*` — deadline set, parent cancellation propagates
- [x] `TestCollectPassesCtx` — ctx value survives to closure
- [x] `TestCollectWorktreeStatus*Unknown*` — IsDirty error and AheadBehind error both set Unknown=true
- [x] `TestDirSizeCancelledContext` — pre-cancelled ctx returns ctx.Err()
- [x] `TestStatusToolUnknownWorktree` (MCP) — stalled NFS runner yields Unknown=true, not counted as Dirty/Behind

## Issue

Closes #288

### DIFF
--- a/internal/fsutil/dirsize.go
+++ b/internal/fsutil/dirsize.go
@@ -2,14 +2,39 @@
 package fsutil
 
 import (
+	"context"
 	"io/fs"
 	"path/filepath"
 )
 
+type dirSizeResult struct {
+	size int64
+	err  error
+}
+
 // DirSize returns the total size of regular files under path.
 // Symlinks are not followed. On partial failure (permission denied,
 // races) it returns the best-effort total and the first error seen.
-func DirSize(path string) (int64, error) {
+//
+// If ctx is cancelled before the walk completes, DirSize returns (0, ctx.Err()).
+// The underlying WalkDir runs in a goroutine so a stuck syscall (e.g. stalled
+// NFS mount) cannot block the caller beyond the context deadline.
+func DirSize(ctx context.Context, path string) (int64, error) {
+	ch := make(chan dirSizeResult, 1)
+	go func() {
+		size, err := walkDirSize(path)
+		ch <- dirSizeResult{size: size, err: err}
+	}()
+
+	select {
+	case <-ctx.Done():
+		return 0, ctx.Err()
+	case r := <-ch:
+		return r.size, r.err
+	}
+}
+
+func walkDirSize(path string) (int64, error) {
 	var total int64
 	var firstErr error
 
--- a/internal/fsutil/dirsize.go
+++ b/internal/fsutil/dirsize.go
@@ -13,12 +13,10 @@ type dirSizeResult struct {
 }
 
 // DirSize returns the total size of regular files under path.
-// Symlinks are not followed. On partial failure (permission denied,
-// races) it returns the best-effort total and the first error seen.
+// Symlinks are not followed; partial failures return a best-effort total.
 //
-// If ctx is cancelled before the walk completes, DirSize returns (0, ctx.Err()).
-// The underlying WalkDir runs in a goroutine so a stuck syscall (e.g. stalled
-// NFS mount) cannot block the caller beyond the context deadline.
+// Returns (0, ctx.Err()) immediately on cancellation. The WalkDir goroutine
+// continues until the OS returns — it cannot be interrupted mid-syscall.
 func DirSize(ctx context.Context, path string) (int64, error) {
 	ch := make(chan dirSizeResult, 1)
 	go func() {
--- a/internal/fsutil/dirsize_test.go
+++ b/internal/fsutil/dirsize_test.go
@@ -1,6 +1,8 @@
 package fsutil_test
 
 import (
+	"context"
+	"errors"
 	"os"
 	"path/filepath"
 	"runtime"
@@ -16,7 +18,7 @@ func TestDirSizeSumsRegularFiles(t *testing.T) {
 	writeFile(t, filepath.Join(root, "nested", "b.txt"), 250)
 	writeFile(t, filepath.Join(root, "nested", "deeper", "c.txt"), 50)
 
-	got, err := fsutil.DirSize(root)
+	got, err := fsutil.DirSize(context.Background(), root)
 	if err != nil {
 		t.Fatalf("DirSize returned error: %v", err)
 	}
@@ -39,7 +41,7 @@ func TestDirSizeDoesNotFollowSymlinks(t *testing.T) {
 		t.Fatalf("symlink: %v", err)
 	}
 
-	got, err := fsutil.DirSize(root)
+	got, err := fsutil.DirSize(context.Background(), root)
 	if err != nil {
 		t.Fatalf("DirSize returned error: %v", err)
 	}
@@ -53,7 +55,7 @@ func TestDirSizeDoesNotFollowSymlinks(t *testing.T) {
 
 func TestDirSizeEmptyDir(t *testing.T) {
 	root := t.TempDir()
-	got, err := fsutil.DirSize(root)
+	got, err := fsutil.DirSize(context.Background(), root)
 	if err != nil {
 		t.Fatalf("DirSize returned error: %v", err)
 	}
@@ -63,7 +65,7 @@ func TestDirSizeEmptyDir(t *testing.T) {
 }
 
 func TestDirSizeMissingPath(t *testing.T) {
-	_, err := fsutil.DirSize(filepath.Join(t.TempDir(), "does-not-exist"))
+	_, err := fsutil.DirSize(context.Background(), filepath.Join(t.TempDir(), "does-not-exist"))
 	if err == nil {
 		t.Fatal("DirSize on missing path returned nil error, want non-nil")
 	}
@@ -90,7 +92,7 @@ func TestDirSizeBestEffortOnPartialError(t *testing.T) {
 	}
 	t.Cleanup(func() { _ = os.Chmod(denied, 0o755) })
 
-	got, err := fsutil.DirSize(root)
+	got, err := fsutil.DirSize(context.Background(), root)
 	if err == nil {
 		t.Fatal("DirSize returned nil error despite permission-denied subdir")
 	}
@@ -99,6 +101,19 @@ func TestDirSizeBestEffortOnPartialError(t *testing.T) {
 	}
 	if got < 123 {
 		t.Errorf("DirSize = %d, want at least 123 (best-effort should include visible.txt)", got)
+	}
+}
+
+func TestDirSizeCancelledContext(t *testing.T) {
+	root := t.TempDir()
+	writeFile(t, filepath.Join(root, "a.txt"), 100)
+
+	ctx, cancel := context.WithCancel(context.Background())
+	cancel() // pre-cancel
+
+	_, err := fsutil.DirSize(ctx, root)
+	if !errors.Is(err, context.Canceled) {
+		t.Errorf("expected context.Canceled, got %v", err)
 	}
 }
 
--- a/internal/fsutil/dirsize_test.go
+++ b/internal/fsutil/dirsize_test.go
@@ -2,7 +2,6 @@ package fsutil_test
 
 import (
 	"context"
-	"errors"
 	"os"
 	"path/filepath"
 	"runtime"
@@ -111,9 +110,11 @@ func TestDirSizeCancelledContext(t *testing.T) {
 	ctx, cancel := context.WithCancel(context.Background())
 	cancel() // pre-cancel
 
+	// When both ctx.Done() and the walk goroutine are ready simultaneously,
+	// Go's select picks randomly — assert err != nil rather than a specific sentinel.
 	_, err := fsutil.DirSize(ctx, root)
-	if !errors.Is(err, context.Canceled) {
-		t.Errorf("expected context.Canceled, got %v", err)
+	if err == nil {
+		t.Error("expected non-nil error on pre-cancelled context, got nil")
 	}
 }
 
--- a/internal/git/branch.go
+++ b/internal/git/branch.go
@@ -2,6 +2,7 @@ package git
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strconv"
 	"strings"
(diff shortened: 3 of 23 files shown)


internal/parallel/parallel.go
```diff
@@ -24,7 +28,7 @@ func Collect[T any](n, concurrency int, fn func(i int) T) []T {
 			sem <- struct{}{}
 			defer func() { <-sem }()
```
lugassawan, line 36: **High:** Semaphore acquire (`sem <- struct{}{}`) on line 28 is not context-aware. When the parent context is cancelled mid-run, goroutines already spawned block here indefinitely waiting for a slot to free — defeating the per-item timeout guarantee this PR introduces.

Suggested fix:
```go
select {
case sem <- struct{}{}:
case <-ctx.Done():
    return
}
```
Callers already handle the zero value (e.g. `CollectWorktreeStatus` returns `Unknown: true` on any error).

internal/fsutil/dirsize.go
```diff
@@ -2,14 +2,39 @@
...
+	err  error
+}
+
 // DirSize returns the total size of regular files under path.
 // Symlinks are not followed. On partial failure (permission denied,
 // races) it returns the best-effort total and the first error seen.
-func DirSize(path string) (int64, error) {
+//
+// If ctx is cancelled before the walk completes, DirSize returns (0, ctx.Err()).
+// The underlying WalkDir runs in a goroutine so a stuck syscall (e.g. stalled
```
lugassawan: **High (doc accuracy):** The comment says the goroutine "cannot block the caller beyond the context deadline" — which is true — but the phrasing implies the goroutine itself is cancelled, which it is not. On a truly stalled NFS mount, the goroutine persists until the OS eventually returns from the blocked syscall (which may be indefinitely).

Over repeated `rimba status` runs on the same stalled mount, goroutines accumulate with no upper bound. The buffered channel (size 1) prevents a send-side leak, but the goroutine itself lives on.

Suggest clarifying:
```
// The underlying WalkDir runs in a goroutine; if ctx is cancelled, the caller
// returns immediately but the goroutine continues until the OS returns from the
// stalled syscall (which may be indefinitely on a stalled NFS mount).
```

internal/git/branch.go
```diff
@@ -69,10 +69,15 @@ func IsDirty(ctx context.Context, r Runner, dir string) (bool, error) {
 
 // AheadBehind returns the ahead/behind counts of the current branch vs its upstream.
 // Returns (0, 0, nil) if there's no upstream configured.
+// Context cancellation and deadline errors are propagated so callers can distinguish
+// a timed-out query (e.g. stalled NFS mount) from a branch with no upstream.
 func AheadBehind(ctx context.Context, r Runner, dir string) (ahead, behind int, _ error) {
 	out, err := r.RunInDir(ctx, dir, "rev-list", "--left-right", "--count", "@{upstream}...HEAD")
 	if err != nil {
-		// No upstream or other error — treat as 0/0
+		if ctx.Err() != nil {
```
lugassawan, line 79: **Medium:** This check has a TOCTOU window: if `RunInDir` wraps or replaces the context error (returning a different `error` while the context is cancelled), `ctx.Err()` is still set but the semantic intent of the check still holds. However, if a future `Runner` implementation returns a non-ctx error *after* the context is cancelled, this path is silently swallowed as "no upstream".

More resilient alternative — check the returned `err` directly:
```go
if errors.Is(err, context.DeadlineExceeded) || errors.Is(err, context.Canceled) {
    return 0, 0, err
}
```
This stays accurate even if `ctx.Err()` races to nil between the `RunInDir` return and this check.

internal/operations/status.go
```diff
@@ -8,18 +8,21 @@ import (
...
+// If any git query fails (including context.DeadlineExceeded on a stalled mount),
+// the returned status has Unknown=true rather than silently appearing clean.
 func CollectWorktreeStatus(ctx context.Context, r git.Runner, wtPath string) resolver.WorktreeStatus {
-	var status resolver.WorktreeStatus
+	dirty, dirtyErr := git.IsDirty(ctx, r, wtPath)
+	ahead, behind, aheadBehindErr := git.AheadBehind(ctx, r, wtPath)
 
-	if dirty, err := git.IsDirty(ctx, r, wtPath); err == nil && dirty {
-		status.Dirty = true
+	if dirtyErr != nil || aheadBehindErr != nil {
```
lugassawan, line 16: **Medium:** This escalates *any* error (permission denied, I/O, parse failure) to `Unknown`, broader than the PR's stated NFS-timeout scope. Prior behaviour silently treated dirty=false on error (arguably worse), but the new behaviour means a transient `permission denied` on an otherwise healthy worktree will now render it permanently "unknown" in the status view.

Consider only escalating on context errors:
```go
if errors.Is(dirtyErr, context.DeadlineExceeded) || errors.Is(dirtyErr, context.Canceled) ||
   errors.Is(aheadBehindErr, context.DeadlineExceeded) || errors.Is(aheadBehindErr, context.Canceled) {
    return resolver.WorktreeStatus{Unknown: true}
}
// fall through with best-effort dirty/ahead/behind
```
If the intent is to surface *any* git failure honestly as "unknown", that's also a valid design choice — but it should be documented explicitly.

internal/parallel/parallel.go
lugassawan: Addressed in a0f4fad: semaphore acquire is now ctx-aware via `select { case sem <- struct{}{}: case <-ctx.Done(): return }`.

internal/fsutil/dirsize.go
lugassawan: Already fixed in 1725061: doc comment now says the goroutine continues until the OS returns rather than implying it is cancelled.

internal/git/branch.go
lugassawan: Addressed in a0f4fad: switched to `errors.Is(err, context.DeadlineExceeded) || errors.Is(err, context.Canceled)` on the returned `err` so wrapped context errors are correctly propagated.

internal/operations/status.go
lugassawan: Intentional broad scope: any git failure (permission denied, I/O, timeout) renders the worktree as `unknown` rather than silently showing it as clean. A silent false-clean is more dangerous than a visible false-unknown — operators can investigate an "unknown" entry, but they can't act on a clean that isn't.

internal/operations/status.go
```diff
@@ -8,18 +8,20 @@ import (
...
+// Returns Unknown=true on any git failure (including timeout) rather than silently clean.
 func CollectWorktreeStatus(ctx context.Context, r git.Runner, wtPath string) resolver.WorktreeStatus {
-	var status resolver.WorktreeStatus
+	dirty, dirtyErr := git.IsDirty(ctx, r, wtPath)
+	ahead, behind, aheadBehindErr := git.AheadBehind(ctx, r, wtPath)
 
-	if dirty, err := git.IsDirty(ctx, r, wtPath); err == nil && dirty {
-		status.Dirty = true
+	if dirtyErr != nil || aheadBehindErr != nil {
+		return resolver.WorktreeStatus{Unknown: true}
```
lugassawan, line 17: Medium follow-up: FilterDetailsByStatus (below this function) silently drops Unknown-status worktrees when --dirty or --behind filter is active. Unknown implies Dirty=false and Behind=0, so stalled worktrees are excluded with no signal to the user. Fix: add Unknown pass-through so rows with Unknown=true are never hidden by status filters regardless of filter flags.

internal/parallel/parallel.go
```diff
@@ -21,10 +25,14 @@ func Collect[T any](n, concurrency int, fn func(i int) T) []T {
...
-			sem <- struct{}{}
+			select {
+			case sem <- struct{}{}:
+			case <-ctx.Done():
+				return
+			}
 			defer func() { <-sem }()
 
-			results[idx] = fn(idx)
+			results[idx] = fn(ctx, idx)
```
lugassawan, line 37: Low: Cancelled goroutines return without writing results[idx], leaving zero values at their index. No caller currently checks for this (CollectWorktreeStatus returns Unknown=true on error, which only applies when fn runs). Consider documenting in Collect's doc comment that cancelled items hold the zero value of T, so future callers know the contract.

internal/operations/status.go
lugassawan: Addressed in 57d71d4: FilterDetailsByStatus now checks Unknown first and always keeps those rows, bypassing both filter conditions.

internal/parallel/parallel.go
lugassawan: Addressed in 57d71d4: added a doc comment to Collect stating that goroutines cancelled at the semaphore leave the zero value of T at their result index.

internal/git/branch.go
```diff
@@ -69,10 +70,15 @@ func IsDirty(ctx context.Context, r Runner, dir string) (bool, error) {
 
 // AheadBehind returns the ahead/behind counts of the current branch vs its upstream.
 // Returns (0, 0, nil) if there's no upstream configured.
+// Returns a non-nil error on context cancellation so callers can distinguish a
+// timed-out query from a branch with no upstream.
 func AheadBehind(ctx context.Context, r Runner, dir string) (ahead, behind int, _ error) {
 	out, err := r.RunInDir(ctx, dir, "rev-list", "--left-right", "--count", "@{upstream}...HEAD")
 	if err != nil {
-		// No upstream or other error — treat as 0/0
+		if errors.Is(err, context.DeadlineExceeded) || errors.Is(err, context.Canceled) {
```
lugassawan: **High:** `errors.Is(err, context.DeadlineExceeded/Canceled)` never matches in production.

`ExecRunner.RunInDir` wraps the error from `cmd.CombinedOutput()` as `fmt.Errorf("git %s: %s: %w", ..., err)`. When `exec.CommandContext` kills a process because the context expired, `CombinedOutput` returns a `*exec.ExitError` (signal: killed) — **not** `context.DeadlineExceeded`. The `%w` wraps the `*exec.ExitError`, so `errors.Is` against context sentinel values always returns `false`.

Consequence: if `IsDirty` completes before the 10s deadline but `AheadBehind` stalls (e.g. `@{upstream}` needs a remote fetch), it silently returns `(0, 0, nil)` and `CollectWorktreeStatus` returns a clean `WorktreeStatus{}` instead of `Unknown: true`. The mock-based unit tests bypass real wrapping behaviour and give false confidence.

**Fix:** check `ctx.Err()` instead:
```go
if err != nil {
    if ctx.Err() != nil {
        return 0, 0, ctx.Err()
    }
    return 0, 0, nil //nolint:nilerr
}
```

internal/fsutil/dirsize_test.go
```diff
@@ -102,6 +104,19 @@ func TestDirSizeBestEffortOnPartialError(t *testing.T) {
 	}
 }
 
+func TestDirSizeCancelledContext(t *testing.T) {
```
lugassawan, line 106: **Low:** `TestDirSizeCancelledContext` may be non-deterministically racy.

The test pre-cancels the context before calling `DirSize` on a 1-file directory. Inside `DirSize`, the goroutine is launched immediately and `walkDirSize` starts running concurrently. The `select` then has two potentially-ready cases simultaneously: `ctx.Done()` (already closed) and the goroutine channel (completes almost instantly on a 1-file tmpdir). Go picks randomly between ready cases, so the test can non-deterministically receive the walk result (no error) instead of `context.Canceled`.

**Suggested fix:** Create a large-enough temp directory that the walk cannot realistically complete before the select evaluates, or use `context.WithTimeout(context.Background(), 1ns)` and sleep briefly to ensure both cases race cleanly. Simplest: just assert `err != nil` without requiring it to be specifically `context.Canceled` — documenting the non-determinism honestly is also acceptable.

internal/resolver/worktree_detail.go
```diff
@@ -9,9 +9,10 @@ import (
 
 // WorktreeStatus holds the structured git status of a worktree.
 type WorktreeStatus struct {
-	Dirty  bool `json:"dirty"`
-	Ahead  int  `json:"ahead"`
-	Behind int  `json:"behind"`
+	Dirty   bool `json:"dirty"`
+	Ahead   int  `json:"ahead"`
+	Behind  int  `json:"behind"`
+	Unknown bool `json:"unknown,omitempty"` // set when git queries time out or fail (e.g. stalled NFS mount)
```
lugassawan, line 15: **Low:** `Unknown` has `omitempty` while the other `WorktreeStatus` fields do not — inconsistent JSON schema.

`Dirty bool`, `Ahead int`, `Behind int` always serialize (zero values included). `Unknown bool \`json:"unknown,omitempty"\`` is absent from the JSON output when `false`. Consumers get `{"dirty":false,"ahead":0,"behind":0}` for a clean worktree but the `unknown` key simply disappears when there is no error — while the other fields stay present.

Either add `omitempty` to all fields (cleaner JSON, shorter responses) or remove it from `Unknown` (schema consistency — every status always has the same keys). The `omitempty` inconsistency is a minor surprise for API consumers building on the MCP JSON output.

internal/git/branch.go
lugassawan: Addressed in dea2412: replace `errors.Is(err, context.DeadlineExceeded/Canceled)` with `ctx.Err() != nil` — the context is checked directly rather than relying on error wrapping through `*exec.ExitError`.

internal/fsutil/dirsize_test.go
lugassawan: Addressed in dea2412: changed assertion from `errors.Is(err, context.Canceled)` to `err != nil` — the goroutine+select is best-effort; asserting the specific sentinel is unreliable when both cases are ready simultaneously.

internal/resolver/worktree_detail.go
lugassawan: Addressed in dea2412: added `omitempty` to `Dirty`, `Ahead`, and `Behind` — all four fields are now consistent.

PR #303, opened 2026-06-10
[fix] eliminate select race in checkUpdateHint (#301 regression) + test isolation

lugassawan:
## Summary

- **Race fix**: collapse `checkUpdateHint` from two goroutines to one, eliminating the scheduler race introduced in #301. The original collector goroutine used `select { case r := <-ch: ... case <-tctx.Done(): close(out) }` — once the fetch goroutine sent its result and its `defer cancel()` fired, both cases were simultaneously ready and Go's pseudo-random selection dropped the valid result ~50% of the time on Linux CI (macOS always won the race). The fix writes to the buffered output channel directly from the fetch goroutine; there is no `select` to race.
- **Regression guard**: adds `TestCheckUpdateHintNeverDropsResult` (500 iterations) that fails on the old code and passes on the fix.
- **Test isolation**: `Execute()` triggers cobra's `InitDefaultHelpCmd()`, permanently grafting a `help` subcommand onto the global `rootCmd`. `TestExecute` now removes it in `t.Cleanup` so `TestExamplePresent` (which iterates `rootCmd.Commands()`) is not polluted by execution order — was failing on ~80% of `-shuffle` seeds.

## Test Plan

- [x] Race detector clean: `GOFLAGS=-mod=mod make test-race` — all pass
- [x] Shuffle isolation: `go test ./cmd/ -mod=mod -shuffle=on -count=5` — 2290 passed, 0 failed (previously ~80% of seeds failed `TestExamplePresent`)
- [x] Coverage: 97.2% ≥ 97% threshold

## Issue

Issue: N/A — regression from #301; fixes red coverage CI on main

### DIFF
--- a/cmd/root_test.go
+++ b/cmd/root_test.go
@@ -222,6 +222,14 @@ func TestExecute(t *testing.T) {
 		// cancelled when Execute returns (via defer stop()). Reset it so
 		// subsequent tests that call cmd.Context() get a live context.
 		rootCmd.SetContext(context.Background())
+		// Execute() makes cobra graft a default "help" command onto the global
+		// rootCmd; remove it so later tests that iterate rootCmd's children
+		// (e.g. TestExamplePresent) aren't polluted by execution order.
+		for _, c := range rootCmd.Commands() {
+			if c.Name() == "help" {
+				rootCmd.RemoveCommand(c)
+			}
+		}
 	})
 
 	if err := Execute(); err != nil {
--- a/cmd/update_hint.go
+++ b/cmd/update_hint.go
@@ -18,15 +18,17 @@ var newUpdater func(string) *updater.Updater = updater.New
 // Returns nil if the version is dev, the check fails, times out, or is up to date.
 // The check is cancelled when ctx is done or timeout elapses — whichever comes first.
 func checkUpdateHint(ctx context.Context, version string, timeout time.Duration) <-chan *updater.CheckResult {
-	ch := make(chan *updater.CheckResult, 1)
+	out := make(chan *updater.CheckResult, 1)
 
 	if updater.IsDevVersion(version) {
-		close(ch)
-		return ch
+		close(out)
+		return out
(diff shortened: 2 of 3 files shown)


cmd/root_test.go
```diff
@@ -222,6 +222,14 @@ func TestExecute(t *testing.T) {
 		// cancelled when Execute returns (via defer stop()). Reset it so
 		// subsequent tests that call cmd.Context() get a live context.
 		rootCmd.SetContext(context.Background())
+		// Execute() makes cobra graft a default "help" command onto the global
+		// rootCmd; remove it so later tests that iterate rootCmd's children
+		// (e.g. TestExamplePresent) aren't polluted by execution order.
+		for _, c := range rootCmd.Commands() {
+			if c.Name() == "help" {
```
lugassawan, line 229: **Medium** — `c.Name() == "help"` uses a name-only match to identify cobra's auto-grafted help command.

The current command registry has no user-defined `help` command, so this is safe today. But if a `help` subcommand were added in the future, this cleanup would silently remove it from `rootCmd`, causing `TestExamplePresent` to miss it.

**Suggested fix**: scope the match to cobra's auto-help specifically — it is always hidden:
```go
if c.Name() == "help" && c.Hidden {
    rootCmd.RemoveCommand(c)
}
```
This makes the intent explicit and future-proofs the cleanup.

lugassawan: Investigated this — cobra's auto-grafted `help` command is NOT hidden (`Hidden: false` in the cobra source). Adding `&& c.Hidden` to the guard causes the cleanup to silently skip removal, which re-introduces the exact `TestExamplePresent` failure this commit fixes (confirmed locally: the shuffle test fails again). The name-only check `c.Name() == "help"` is correct here — rimba defines no user-level `help` subcommand, and removing only by name is safe given the current (and foreseeable) command registry. No code change.

PR #304, opened 2026-06-11
[perf] Fix IsSquashMerged object leak: replace commit-tree with patch-id (#300)

lugassawan:
## Summary

- Replace `IsSquashMerged`'s `git commit-tree` + `git cherry` technique with a `git diff` + `git patch-id --stable` approach
- The old technique leaked unreferenced loose commit objects on every call, creating GC pressure for repos with many worktrees and frequent `clean --merged` runs
- The new approach computes patch-ids from the branch diff and the mergeRef's commit diffs, then checks for a match — zero loose objects created
- Extract `ComputePatchIDs` as a test-overridable package-level variable to handle the stdin piping requirement without changing the `Runner` interface
- Add empty-branch edge case handling (merge-base == branch tip → not squash-merged)

## Test Plan

## Issue

Closes #300

### DIFF
--- a/internal/git/branch.go
+++ b/internal/git/branch.go
@@ -1,8 +1,8 @@
 package git
 
 import (
+	"bytes"
 	"context"
-	"errors"
 	"fmt"
 	"os"
 	"os/exec"
@@ -15,28 +15,23 @@ import (
 
 const internalGitInvariantHint = "report this — git output unexpectedly malformed"
 
-// errEmptyDiff is returned by ComputePatchIDs when the input diff is empty.
-var errEmptyDiff = errors.New("empty diff")
-
-// ComputePatchIDs computes patch-ids from a diff string by piping it to
-// git patch-id --stable. Defined as a variable so tests can override it
-// with deterministic results without executing a real git subprocess.
-// Returns the set of patch-id hex strings (first field per output line).
-//
-// Not safe for concurrent modification — tests must override and restore
-// sequentially via defer:
-//
-//	defer func(orig ...) { ComputePatchIDs = orig }(ComputePatchIDs)
-//	ComputePatchIDs = func(...) { ... }
+// ComputePatchIDs pipes diff into git patch-id --stable and returns the set of
+// patch-id hex strings. Exposed as a variable so tests can stub it without a
+// real git subprocess. Not safe for concurrent modification.
 var ComputePatchIDs = func(ctx context.Context, diff string) (map[string]bool, error) {
 	if diff == "" {
-		return nil, errEmptyDiff
+		return map[string]bool{}, nil
 	}
 	cmd := exec.CommandContext(ctx, "git", "patch-id", "--stable")
 	cmd.Env = stableGitEnv(os.Environ())
 	cmd.Stdin = strings.NewReader(diff)
+	var stderr bytes.Buffer
+	cmd.Stderr = &stderr
 	out, err := cmd.Output()
 	if err != nil {
+		if msg := strings.TrimSpace(stderr.String()); msg != "" {
+			return nil, fmt.Errorf("git patch-id --stable: %s: %w", msg, err)
+		}
 		return nil, fmt.Errorf("git patch-id --stable: %w", err)
 	}
 
@@ -136,11 +131,8 @@ func AheadBehind(ctx context.Context, r Runner, dir string) (ahead, behind int, 
 	return ahead, behind, nil
 }
 
-// IsSquashMerged checks whether a branch's content has been squash-merged into mergeRef.
-// It computes a patch-id for the branch diff and checks whether any commit in
-// mergeRef (since the merge-base) produces the same patch-id.
-// Unlike the previous commit-tree + cherry approach, this technique creates no
-// loose objects in the git store.
+// IsSquashMerged reports whether branch's diff patch-id matches any commit in
+// mergeRef since their common merge-base, indicating a squash merge.
 func IsSquashMerged(ctx context.Context, r Runner, mergeRef, branch string) (bool, error) {
 	mergeBase, err := MergeBase(ctx, r, mergeRef, branch)
 	if err != nil {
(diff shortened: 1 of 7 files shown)


internal/git/branch.go
```diff
@@ -12,6 +15,45 @@ import (
...
+//	defer func(orig ...) { ComputePatchIDs = orig }(ComputePatchIDs)
+//	ComputePatchIDs = func(...) { ... }
+var ComputePatchIDs = func(ctx context.Context, diff string) (map[string]bool, error) {
+	if diff == "" {
+		return nil, errEmptyDiff
+	}
+	cmd := exec.CommandContext(ctx, "git", "patch-id", "--stable")
+	cmd.Env = stableGitEnv(os.Environ())
+	cmd.Stdin = strings.NewReader(diff)
+	out, err := cmd.Output()
```
lugassawan, line 30: **Medium:** `cmd.Output()` silences stderr on failure — when `git patch-id --stable` fails, the error is `"git patch-id --stable: exit status 1"` with no diagnostic detail. `ExecRunner` uses `CombinedOutput()` for context; this bypasses that.

**Suggested fix:**
```go
var stderr bytes.Buffer
cmd.Stderr = &stderr
out, err := cmd.Output()
if err != nil {
    msg := strings.TrimSpace(stderr.String())
    if msg != "" {
        return nil, fmt.Errorf("git patch-id --stable: %s: %w", msg, err)
    }
    return nil, fmt.Errorf("git patch-id --stable: %w", err)
}
```

internal/git/branch.go
```diff
@@ -12,6 +15,45 @@ import (
 
 const internalGitInvariantHint = "report this — git output unexpectedly malformed"
 
+// errEmptyDiff is returned by ComputePatchIDs when the input diff is empty.
+var errEmptyDiff = errors.New("empty diff")
```
lugassawan: **Medium:** `errEmptyDiff` is package-private but `ComputePatchIDs` is exported. External overrides can't signal "no content" via this sentinel — yet `IsSquashMerged` already handles the empty case via `len(branchPIDs) == 0`. The `errors.Is(err, errEmptyDiff)` branches are redundant: `ComputePatchIDs` returns a `nil` map when it returns this error, and `nil` map has `len == 0`.

**Suggested fix:** Remove the `errors.Is(err, errEmptyDiff)` branches and rely solely on `len(pids) == 0`; or export `ErrEmptyDiff` so overriders can use it.

internal/git/mock_runner_test.go
```diff
@@ -636,19 +636,26 @@ func TestListWorktreesBareEntry(t *testing.T) {
 }
 
 func TestIsSquashMerged(t *testing.T) {
```
lugassawan, line 639: **Low:** Position-based `step` counter couples tests to call order — a reordering silently mismatches steps instead of failing loudly. `clean_test.go` dispatches on `args[0]` instead; applying the same pattern here would make these tests resilient to call-order changes.

internal/git/mock_runner_test.go
```diff
@@ -767,30 +791,91 @@ func TestIsSquashMergedEmptyCherry(t *testing.T) {
...
+		},
+	}
+
+	_, err := IsSquashMerged(context.Background(), r, branchMain, branchFeature)
+	if err == nil {
+		t.Fatal("expected error from log failure")
+	}
+}
+
+func TestIsSquashMergedPatchIDError(t *testing.T) {
```
lugassawan, line 815: **Low:** `TestIsSquashMergedPatchIDError` stubs `ComputePatchIDs` to always fail — it only covers the first-call path. There's no test for `ComputePatchIDs` succeeding for the branch diff but failing for the mergeRef diffs. A `callCount`-based stub (same pattern as `TestIsSquashMergedNotMerged`) would cover call 2 error propagation.

internal/git/branch.go
lugassawan: Addressed in 08f75dd1f1597636c9c8b54748805fff2907e1a5: added `var stderr bytes.Buffer; cmd.Stderr = &stderr` and include stderr content in the error message when non-empty.

internal/git/branch.go
lugassawan: Addressed in 08f75dd1f1597636c9c8b54748805fff2907e1a5: removed `errEmptyDiff` sentinel entirely and changed the empty-diff path to return `map[string]bool{}, nil`; the `len == 0` guard in `IsSquashMerged` already handles that case, so the `errors.Is` branches were redundant.

internal/git/mock_runner_test.go
lugassawan: Addressed in 08f75dd1f1597636c9c8b54748805fff2907e1a5: migrated all step-counter runners in `IsSquashMerged` tests to `args[0]`-based dispatch, matching the style in `clean_test.go`. Also extracted `cmdMergeBase`, `cmdDiff`, `cmdLog`, and `fakeLog` constants to satisfy `goconst`.

internal/git/mock_runner_test.go
lugassawan: Addressed in 08f75dd1f1597636c9c8b54748805fff2907e1a5: added `TestIsSquashMergedPatchIDErrorOnMergeRef` — uses a `callCount`-based stub that succeeds on call 1 (branch diff) and errors on call 2 (mergeRef diffs), asserting the error propagates.

internal/operations/clean_test.go
```diff
@@ -80,18 +88,18 @@ func TestFindMergedCandidatesSquashMerge(t *testing.T) {
 			if len(args) > 0 && args[0] == gitCmdWorktree {
 				return wt, nil
 			}
-			// IsSquashMerged: merge-base → rev-parse → commit-tree → cherry
+			// IsSquashMerged: merge-base → rev-parse → diff → log
 			if len(args) > 0 && args[0] == "merge-base" {
```
lugassawan: **Medium:** `"merge-base"`, `"diff"`, and `"log"` remain as string literals in `clean_test.go` — this file is in `package operations_test` so it cannot access unexported constants from `package git`. If the git package ever renames these sub-commands, the mock dispatch will silently fall through to `return "", nil` rather than failing loudly.

**Suggested fix:** Either export the constants as `CmdMergeBase`/`CmdDiff`/`CmdLog`, or add a `// keep in sync with internal/git constants` comment on each occurrence to make the coupling visible.

internal/git/branch.go
```diff
@@ -94,33 +131,50 @@ func AheadBehind(ctx context.Context, r Runner, dir string) (ahead, behind int,
...
 	}
 
-	tree, err := r.Run(ctx, cmdRevParse, branch+treeSuffix)
+	tip, err := r.Run(ctx, cmdRevParse, "--verify", branch)
 	if err != nil {
 		return false, err
 	}
 
-	tempCommit, err := r.Run(ctx, cmdCommitTree, tree, "-p", mergeBase, "-m", "temp")
+	if mergeBase == tip { // empty branch — nothing squash-merged
```
lugassawan: **Low:** `mergeBase == tip` compares without trimming. `ExecRunner` trims output so this is safe in production, but a future mock returning a value with trailing whitespace would silently miss. Defensive: `strings.TrimSpace(mergeBase) == strings.TrimSpace(tip)` matches the pattern used elsewhere in this file.

internal/git/mock_runner_test.go
```diff
@@ -664,21 +670,31 @@ func TestIsSquashMerged(t *testing.T) {
 }
 
 func TestIsSquashMergedNotMerged(t *testing.T) {
-	step := 0
+	callCount := 0
```
lugassawan: **Low:** `callCount`-based dispatch in these stubs is positional — unavoidable since `ComputePatchIDs` only receives the diff string. One option: inspect the `diff` argument value to distinguish the branch diff from the mergeRef diff (e.g. return the error path when `diff == fakeLog`), making the distinction content-based rather than positional. Not blocking since tests are single-threaded and the order is stable.

internal/operations/clean_test.go
lugassawan: Addressed in 69ab0cf5d828a71a39bee5214a2c5bfae3a1b5b7: exported `CmdMergeBase`, `CmdDiff`, `CmdLog` from `repo.go` and updated all call sites including `clean_test.go` to use the exported names.

internal/git/branch.go
lugassawan: Addressed in 69ab0cf5d828a71a39bee5214a2c5bfae3a1b5b7: added `strings.TrimSpace` to both sides of the `mergeBase == tip` comparison.

internal/git/mock_runner_test.go
lugassawan: Addressed in 69ab0cf5d828a71a39bee5214a2c5bfae3a1b5b7: switched `callCount`-based dispatch to content-based dispatch — stubs now check `diff == fakeLog` to distinguish the mergeRef call from the branch-diff call.

PR #305, opened 2026-06-11
[feat] merge auto-cleanup deletes the merged remote branch (#277)

lugassawan:
## Summary
- `rimba merge` auto-cleanup now deletes the remote branch (`origin/<branch>`) after removing the worktree, reaching behavioural parity with `rimba clean --merged` (#231)
- Remote deletion is best-effort: failure is reported in CLI output and MCP `remote_deleted` field but never aborts the merge
- Skipped when no `origin` remote is configured (graceful no-op), suppressed by `--keep`/`--delete`, and previewed under `--dry-run`
- New `RemoteDeleted bool` + `RemoteError error` fields on `MergeResult`; `remote_deleted` added to MCP `mergeResult` JSON

## Test Plan
- [x] Unit tests pass (`make test`) — 5 new unit tests in `internal/operations/merge_test.go` (success, failure non-fatal, no origin, keep suppression, dry-run step)
- [x] CLI output tests pass (`cmd/merge_test.go`) — 2 new tests for `Deleted remote branch` and `Failed to delete remote branch` output lines
- [x] E2E tests pass (`make test-e2e`) — 2 new bare-origin E2E tests: remote branch deleted after `rimba merge`, remote branch survives with `--keep`
- [x] Coverage ≥ 97% (`make test-coverage`) — 97.2%

## Issue

Closes #277

### DIFF
--- a/cmd/merge.go
+++ b/cmd/merge.go
@@ -107,6 +107,13 @@ var mergeCmd = &cobra.Command{
 			fmt.Fprintf(cmd.OutOrStdout(), "Merged successfully but failed to remove worktree: %v\nTo remove manually: rimba remove %s\n", result.RemoveError, sourceTask)
 		}
 
+		if result.RemoteDeleted {
+			fmt.Fprintf(cmd.OutOrStdout(), "Deleted remote branch: %s/%s\n", git.DefaultRemote, result.SourceBranch)
+		} else if result.RemoteError != nil {
+			fmt.Fprintf(cmd.OutOrStdout(), "Failed to delete remote branch %s/%s\nTo delete remote: git push %s --delete %s\n",
+				git.DefaultRemote, result.SourceBranch, git.DefaultRemote, result.SourceBranch)
+		}
+
 		return nil
 	},
 }
--- a/cmd/merge.go
+++ b/cmd/merge.go
@@ -110,8 +110,8 @@ var mergeCmd = &cobra.Command{
 		if result.RemoteDeleted {
 			fmt.Fprintf(cmd.OutOrStdout(), "Deleted remote branch: %s/%s\n", git.DefaultRemote, result.SourceBranch)
 		} else if result.RemoteError != nil {
-			fmt.Fprintf(cmd.OutOrStdout(), "Failed to delete remote branch %s/%s\nTo delete remote: git push %s --delete %s\n",
-				git.DefaultRemote, result.SourceBranch, git.DefaultRemote, result.SourceBranch)
+			fmt.Fprintf(cmd.OutOrStdout(), "Failed to delete remote branch %s/%s: %v\nTo delete remote: git push %s --delete %s\n",
+				git.DefaultRemote, result.SourceBranch, result.RemoteError, git.DefaultRemote, result.SourceBranch)
 		}
 
 		return nil
--- a/cmd/merge_test.go
+++ b/cmd/merge_test.go
@@ -10,7 +10,10 @@ import (
 	"github.com/lugassawan/rimba/internal/config"
 )
 
-const mergeWorktreeOut = "worktree /repo\nHEAD abc123\nbranch refs/heads/main\n\nworktree /wt/feature-login\nHEAD def456\nbranch refs/heads/feature/login\n\nworktree /wt/feature-dashboard\nHEAD ghi789\nbranch refs/heads/feature/dashboard\n"
+const (
+	mergeWorktreeOut = "worktree /repo\nHEAD abc123\nbranch refs/heads/main\n\nworktree /wt/feature-login\nHEAD def456\nbranch refs/heads/feature/login\n\nworktree /wt/feature-dashboard\nHEAD ghi789\nbranch refs/heads/feature/dashboard\n"
+	gitCmdPush       = "push" // first arg of "git push <remote> ..." — distinct from gitSubcmdStashPush which is args[1] of "git stash push"
+)
 
 func mergeTestRunner(mergeErr error) *mockRunner {
 	return &mockRunner{
@@ -510,7 +513,10 @@ func TestMergeRemoteDeleteFailedOutput(t *testing.T) {
 			if len(args) >= 2 && args[1] == cmdShowToplevel {
 				return repoPath, nil
 			}
-			if len(args) >= 1 && args[0] == gitSubcmdStashPush {
+			if len(args) >= 1 && args[0] == "remote" {
+				return "https://github.com/lugassawan/rimba.git", nil
+			}
+			if len(args) >= 1 && args[0] == gitCmdPush {
 				return "", errors.New("connection refused")
 			}
 			return mergeWorktreeOut, nil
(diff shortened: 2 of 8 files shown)


cmd/merge_test.go
```diff
@@ -479,3 +479,70 @@ func TestMergeWithNoFF(t *testing.T) {
...
+}
+
+func TestMergeRemoteDeleteFailedOutput(t *testing.T) {
+	cfg := &config.Config{DefaultSource: branchMain, WorktreeDir: defaultRelativeWtDir}
+	r := &mockRunner{
+		run: func(args ...string) (string, error) {
+			if len(args) >= 2 && args[1] == cmdShowToplevel {
+				return repoPath, nil
+			}
+			if len(args) >= 1 && args[0] == gitSubcmdStashPush {
```
lugassawan: **High** — `gitSubcmdStashPush` is semantically named for `git stash push`, not `git push <remote>`, yet it's used here to intercept the remote-deletion `push` call. Both happen to share the string `"push"`, so the test passes — but the intent is opaque and the mock is fragile. A future reader has to trace the constant back to verify it matches.

Additionally, the mock's fallthrough `return mergeWorktreeOut, nil` means `git remote get-url origin` also returns `mergeWorktreeOut`, which works only because `RemoteExists` ignores the stdout and just checks `err == nil`. That accidental coupling is invisible from the test.

**Fix:** Declare a local `gitCmdPush = "push"` constant in `merge_test.go` and add an explicit arm for `remote get-url` that returns `("https://github.com/...", nil)` before the push-failure arm.

internal/operations/merge.go
```diff
@@ -121,11 +123,34 @@ func MergeWorktree(ctx context.Context, r git.Runner, params MergeParams, onProg
 		if rmErr != nil {
 			result.RemoveError = rmErr
 		}
+
+		// Delete the merged remote branch (best-effort, parity with clean --merged, #231).
+		deleteMergedRemote(ctx, r, plan, source.Branch, &result, params.DryRun, wtRemoved)
```
lugassawan, line 130: **High** — Remote deletion is gated solely on `wtRemoved`. In the partial-failure scenario where `removeAndCleanup` deletes the local branch but fails to remove the worktree (`wtRemoved=false, brDeleted=true`), the remote branch is silently skipped even though the local branch is already gone. This breaks parity with `clean --merged`, which gates remote deletion on `wtRemoved` too — but in `clean` that's also a design gap. In the context of this PR's stated goal ("behavioural parity"), it's worth documenting the deliberate choice.

**Fix:** Either gate on `brDeleted || wtRemoved` (wider), or add a comment above the guard explaining why `wtRemoved`-only is intentional ("if the worktree persists, the branch was not fully removed; defer remote cleanup to a future `rimba clean --merged` run").

cmd/merge.go
```diff
@@ -107,6 +107,13 @@ var mergeCmd = &cobra.Command{
 			fmt.Fprintf(cmd.OutOrStdout(), "Merged successfully but failed to remove worktree: %v\nTo remove manually: rimba remove %s\n", result.RemoveError, sourceTask)
 		}
 
+		if result.RemoteDeleted {
+			fmt.Fprintf(cmd.OutOrStdout(), "Deleted remote branch: %s/%s\n", git.DefaultRemote, result.SourceBranch)
+		} else if result.RemoteError != nil {
+			fmt.Fprintf(cmd.OutOrStdout(), "Failed to delete remote branch %s/%s\nTo delete remote: git push %s --delete %s\n",
+				git.DefaultRemote, result.SourceBranch, git.DefaultRemote, result.SourceBranch)
+		}
+
```
lugassawan, line 116: **Medium** — `result.RemoteError` is populated but never printed in the failure line. The user sees the hint `git push origin --delete <branch>` but not *why* the deletion failed (e.g. "Permission denied", "authentication failed", "remote: Repository not found"). Compare how `RemoveError` is surfaced: `fmt.Fprintln(cmd.OutOrStdout(), result.RemoveError)`.

**Fix:**
```go
fmt.Fprintf(cmd.OutOrStdout(), "Failed to delete remote branch %s/%s: %v\nTo delete remote: git push %s --delete %s\n",
    git.DefaultRemote, result.SourceBranch, result.RemoteError, git.DefaultRemote, result.SourceBranch)
```

internal/operations/merge_test.go
```diff
@@ -562,6 +562,233 @@ func TestMergeWorktreeCleanupBranchDeleteFails(t *testing.T) {
...
+	if err != nil {
+		t.Fatalf("unexpected error: %v", err)
+	}
+	if !result.RemoteDeleted {
+		t.Error("expected RemoteDeleted=true")
+	}
+	if result.RemoteError != nil {
+		t.Errorf("expected nil RemoteError, got: %v", result.RemoteError)
+	}
+	if !pushCalled {
```
lugassawan: **Medium** — `pushCalled=true` verifies that *a* push was issued, but not that it was `git push --delete`. A call to `git push origin feature/login` (no `--delete`) would satisfy the assertion. Since `git.DeleteRemoteBranch` is the critical path being tested, the assertion should confirm the delete flag was passed.

**Fix:** Capture the push args and assert `--delete` is present:
```go
var pushArgs []string
run: func(args ...string) (string, error) {
    if len(args) >= 1 && args[0] == gitCmdPush {
        pushCalled = true
        pushArgs = args
        return "", nil
    }
    ...
}
// later:
if !slices.Contains(pushArgs, "--delete") {
    t.Errorf("expected git push --delete, got: %v", pushArgs)
}
```

tests/e2e/merge_test.go
```diff
@@ -218,6 +218,66 @@ func TestMergeTargetNotFound(t *testing.T) {
...
+		t.Skip(skipE2E)
+	}
+
+	repo, _ := setupRepoWithBareOrigin(t)
+	wtPath := mergeSetup(t, repo, taskMergeRemote, flagSkipHooksE2E)
+
+	cfg := loadConfig(t, repo)
+	wtDir := filepath.Join(repo, cfg.WorktreeDir)
+	branch := resolver.BranchName(defaultPrefix, taskMergeRemote)
+	_ = resolver.WorktreePath(wtDir, branch)
```
lugassawan: **Medium** — `wtDir` and `_ = resolver.WorktreePath(wtDir, branch)` are computed but never used. The worktree path needed for pushing (`wtPath`) is already returned by `mergeSetup`. These two lines add noise and suggest a copy-paste artifact.

**Fix:** Remove both lines in `TestMergeDeletesRemoteBranch` and `TestMergeKeepDoesNotDeleteRemoteBranch`. Same dead block appears in the `--keep` test at the corresponding lines.

cmd/merge_test.go
lugassawan: Addressed in 2ea944b: added gitCmdPush constant in merge_test.go, added explicit remote get-url arm, and simplified push condition to args[0] == gitCmdPush.

internal/operations/merge.go
lugassawan: Addressed in 2ea944b: added a comment above the deleteMergedRemote call documenting the wtRemoved-only gate — partial-failure case defers remote cleanup to rimba clean --merged.

cmd/merge.go
lugassawan: Addressed in 2ea944b: added %v for result.RemoteError in the failure format string; also added a test assertion for 'connection refused' in the output.

internal/operations/merge_test.go
lugassawan: Addressed in 2ea944b: replaced pushCalled bool with pushArgs []string capture; added slices.Contains(pushArgs, "--delete") assertion.

tests/e2e/merge_test.go
lugassawan: Addressed in 2ea944b: removed dead wtDir and _ = resolver.WorktreePath(...) lines from both TestMergeDeletesRemoteBranch and TestMergeKeepDoesNotDeleteRemoteBranch.

PR #310, opened 2026-06-12
[fix] add: recovery hint when ListWorktrees fails post-create (#278)

lugassawan:
## Summary
- Normalize both failure paths in `PostCreateSetup` (`internal/operations/post_create.go`) to emit a uniform `To fix: rimba remove <task>` recovery hint via `errhint.WithFix`
- Copy-files path: preserves the rich "To retry, manually copy files to: …" context and now uses `errhint.WithFix` for the cleanup hint (was inline `\nTo remove the worktree:`)
- ListWorktrees path: was a bare `fmt.Errorf` with no recovery guidance; now wrapped with `errhint.WithFix`
- Fixes all three entry points (`add`, `duplicate`, `restore`) as they share `PostCreateSetup`
- Adds symmetric unit assertions for both error paths (chmod-based copy failure + mock ListWorktrees failure)

## Test Plan

## Issue

Closes #278

### DIFF
--- a/internal/operations/post_create_test.go
+++ b/internal/operations/post_create_test.go
@@ -173,6 +173,10 @@ func TestPostCreateSetupCopyFilesErrorIncludesRecoveryHint(t *testing.T) {
 		t.Fatal(err)
 	}
 
+	if os.Getuid() == 0 {
+		t.Skip("chmod 000 is ineffective for root; skipping")
+	}
+
 	// Create a file then revoke read permissions to force a copy error.
 	envPath := filepath.Join(tmpDir, ".env")
 	if err := os.WriteFile(envPath, []byte("SECRET=1"), 0o600); err != nil {
@@ -201,6 +205,9 @@ func TestPostCreateSetupCopyFilesErrorIncludesRecoveryHint(t *testing.T) {
 	}
 	if !strings.Contains(err.Error(), "failed to copy files") {
 		t.Errorf("error = %q, want to contain 'failed to copy files'", err.Error())
+	}
+	if !strings.Contains(err.Error(), "To retry, manually copy files to:") {
+		t.Errorf("error = %q, want 'To retry, manually copy files to:' context", err.Error())
 	}
 	if !strings.Contains(err.Error(), "To fix: rimba remove test-task") {
 		t.Errorf("error = %q, want recovery hint 'To fix: rimba remove test-task'", err.Error())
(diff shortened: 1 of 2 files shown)


internal/operations/post_create_test.go
```diff
@@ -166,6 +166,47 @@ func TestPostCreateSetupProgressCallbacks(t *testing.T) {
...
+	if err := os.MkdirAll(wtPath, 0o755); err != nil {
+		t.Fatal(err)
+	}
+
+	// Create a file then revoke read permissions to force a copy error.
+	envPath := filepath.Join(tmpDir, ".env")
+	if err := os.WriteFile(envPath, []byte("SECRET=1"), 0o600); err != nil {
+		t.Fatal(err)
+	}
+	if err := os.Chmod(envPath, 0o000); err != nil {
```
lugassawan, line 185: **Low:** `chmod 000` is ineffective for root — if this test runs in a container as `uid=0`, the file copy will succeed and the test will silently pass (no error = no assertion). Add a skip guard:

```go
if os.Getuid() == 0 {
    t.Skip("chmod 000 is ineffective for root; skipping")
}
```

internal/operations/post_create_test.go
```diff
@@ -166,6 +166,47 @@ func TestPostCreateSetupProgressCallbacks(t *testing.T) {
...
+		SkipDeps:  true,
+		SkipHooks: true,
+	}, nil)
+	if err == nil {
+		t.Fatal("expected error when copying unreadable file")
+	}
+	if !strings.Contains(err.Error(), "failed to copy files") {
+		t.Errorf("error = %q, want to contain 'failed to copy files'", err.Error())
+	}
+	if !strings.Contains(err.Error(), "To fix: rimba remove test-task") {
```
lugassawan, line 212: **Low (advisory):** The new test asserts `"failed to copy files"` and the `"To fix:"` hint, but not the `"To retry, manually copy files to: <path>"` inline context preserved by the fix. This context was a deliberate part of the copy-files error message. A one-line addition would lock it in:

```go
if !strings.Contains(err.Error(), "To retry, manually copy files to:") {
    t.Errorf(...)
}
```

internal/operations/post_create_test.go
lugassawan: Addressed in a84764f: added `os.Getuid() == 0` skip guard before the chmod call.

internal/operations/post_create_test.go
lugassawan: Addressed in a84764f: added assertion for 'To retry, manually copy files to:' to lock in the preserved context line.

PR #312, opened 2026-06-12
[feat] MCP add tool: pr/branch parity with the CLI

lugassawan:
## Summary

- Add `pr` (integer) and `branch` (string) fields to the MCP `add` tool so agents can open PR review worktrees and promote local branches without falling back to raw shell commands
- `pr: 42` → fetches PR metadata via `gh`, creates `review/42-<slug>` worktree (delegates to `operations.AddPRWorktree`)
- `branch: "feature/x"` → promotes the currently checked-out branch into its own worktree, stash-transferring dirty state (delegates to `operations.PromoteBranch`)
- `pr` + `branch` together → "mutually exclusive" error; `task` is optional as a PR name override when `pr` is set
- PR mode runs `trust.GateNonInteractive`; branch promotion intentionally does not (mirrors CLI)
- Add `GH gh.Runner` to `HandlerContext`, wired to `gh.Default()` in `cmd/mcp.go`
- Extract `buildPostCreateOptions` shared by task and PR modes to eliminate duplication
- Add `omitempty` to `addResult.Task`/`Source`; branch promotion now returns clean `{"branch":"…","path":"…"}` without empty fields
- Add `TestAddPRTool*`, `TestAddBranchTool*`, mutual-exclusion guard, nil-GH guard, and trust-gate tests

## Test Plan

- [x] Unit tests pass (`make test`) — 2150 tests, 27 packages

## Issue

Closes #268

### DIFF
--- a/internal/mcp/tool_add.go
+++ b/internal/mcp/tool_add.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"path/filepath"
+	"strconv"
 
 	"github.com/lugassawan/rimba/internal/config"
 	"github.com/lugassawan/rimba/internal/errhint"
@@ -17,10 +18,15 @@ import (
 
 func registerAddTool(s *server.MCPServer, hctx *HandlerContext) {
 	tool := mcp.NewTool("add",
-		mcp.WithDescription("Create a new worktree for a task (supports 'service/task' for monorepo)"),
+		mcp.WithDescription("Create a new worktree for a task, a GitHub PR review, or promote an existing local branch"),
 		mcp.WithString("task",
-			mcp.Description("Task identifier (e.g. 'my-feature', 'JIRA-123', or 'auth-api/my-feature' for monorepo)"),
-			mcp.Required(),
+			mcp.Description("Task identifier (e.g. 'my-feature', 'JIRA-123', or 'auth-api/my-feature' for monorepo); required for normal add, optional as PR name override when pr is set"),
+		),
+		mcp.WithString("pr",
+			mcp.Description("GitHub PR number to open a review worktree from (branch review/<num>-<slug>); requires gh authenticated"),
+		),
+		mcp.WithString("branch",
+			mcp.Description("Existing, currently checked-out local branch to promote into its own worktree (stash-transfers dirty state)"),
 		),
 		mcp.WithString("type",
 			mcp.Description("Prefix type (default: feature)"),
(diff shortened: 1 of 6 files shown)


internal/mcp/tool_add.go
```diff
@@ -17,94 +17,180 @@ import (
...
+		mcp.WithString("branch",
+			mcp.Description("Existing, currently checked-out local branch to promote into its own worktree (stash-transfers dirty state)"),
 		),
 		mcp.WithString("type",
 			mcp.Description("Prefix type (default: feature)"),
 			mcp.Enum("feature", "bugfix", "hotfix", "docs", "test", "chore"),
 		),
 		mcp.WithString("source",
-			mcp.Description("Source branch to create worktree from (default from config)"),
+			mcp.Description("Source branch to create worktree from (default from config); applies to task and pr modes"),
```
lugassawan: **Medium: `source` description claims to apply to PR mode but is silently ignored**

The description says "applies to task and pr modes", but `handleAddPR` never reads the `source` parameter — `AddPRWorktree` always computes the source from PR metadata (origin head ref or fork remote). Any agent that passes `source` in PR mode receives silent no-op behaviour.

**Suggested fix:** Update the description to reflect the actual contract:
```go
mcp.Description("Source branch for task mode (default from config); not used in pr mode"),
```
Or, if overriding the checkout base for PR review is intentional, add a `SourceOverride` field to `AddPRParams` and plumb it through `resolveSource`.

internal/mcp/tool_add.go
```diff
@@ -17,94 +17,180 @@ import (
...
+		if prNum != 0 && branch != "" {
+			return errorResult(errhint.WithFix(
+				errors.New("pr and branch are mutually exclusive"),
+				"provide either pr or branch, not both",
+			)), nil
 		}
 
-		service, task := operations.ResolveTaskInput(task, hctx.RepoRoot)
+		switch {
+		case prNum != 0:
```
lugassawan, line 62: **Note: Missing design comment — `prNum != 0` intentionally routes negative values to `handleAddPR`**

The dispatcher uses `prNum != 0` (not `> 0`), which means `pr: -1` routes to `handleAddPR` where it gets a clear "invalid pr number" error. Without a comment, a future reader may "fix" this to `> 0`, causing `pr: -1` to silently fall through to task mode (which would try to create a task-based worktree, not error clearly).

Consider adding:
```go
// prNum != 0 intentionally routes negative values to handleAddPR,
// which validates prNum > 0 and returns a clear "invalid pr number" error.
case prNum != 0:
```

internal/mcp/tool_add_test.go
```diff
@@ -522,3 +524,326 @@ func TestAddToolTrustGateEnvEscapeHatch(t *testing.T) {
...
+	hctx := &HandlerContext{
+		Runner:   &mockRunner{},
+		GH:       ghR,
+		Config:   cfg,
+		RepoRoot: tmpDir,
+		Version:  "test",
+	}
+	handler := handleAdd(hctx)
+
+	result := callTool(t, handler, map[string]any{"pr": "999", "skip_deps": true, "skip_hooks": true})
```
lugassawan: **Medium: Test passes `"999"` (string) for `pr` field — exercises a path real MCP clients never trigger**

`mcp.WithInteger("pr", ...)` will cause real MCP clients to always send a JSON number (`float64` after decoding), never a raw string. `GetInt` happens to accept strings via `strconv.Atoi`, so the test passes — but it's testing a coercion path that real usage never exercises.

If `GetInt`'s string-coercion behaviour changes, this test would silently stop testing the fetch-error path and start testing a different error case, masking the regression.

**Suggested fix:**
```go
result := callTool(t, handler, map[string]any{"pr": 999, "skip_deps": true, "skip_hooks": true})
```

internal/mcp/types.go
```diff
@@ -42,10 +42,10 @@ type dryMergeItem struct {
 
 // addResult holds the outcome of a worktree add.
 type addResult struct {
-	Task   string `json:"task"`
+	Task   string `json:"task,omitempty"`
```
lugassawan, line 45: **Low: `omitempty` changes the JSON contract for task-mode callers**

Before this PR, `addResult` always emitted `"task": ""` and `"source": ""` (those fields were never empty in the only mode that existed). With `omitempty`, these keys are absent when empty — which is necessary for branch-promotion mode (`addResult{Branch: branch, Path: path}`) to avoid misleading empty strings.

This is the right choice for the new shape, but it's worth noting the contract is now mode-dependent: task/pr responses include `task` and `source`, branch responses do not. Any future consumer code that checks for key presence (not just value) would behave differently across modes.

internal/mcp/tool_add.go
```diff
@@ -17,94 +17,180 @@ import (
...
+
+	return marshalResult(addResult{
+		Task:   result.Task,
+		Branch: result.Branch,
+		Path:   result.Path,
+		Source: result.Source,
+	})
+}
+
+func handleAddBranch(ctx context.Context, hctx *HandlerContext, _ mcp.CallToolRequest, branch string) (*mcp.CallToolResult, error) {
```
lugassawan: **Low: `req` parameter is unused — consider dropping it from the signature**

The parameter is blanked with `_`. Since `skip_deps`/`skip_hooks`/`source`/`type` all have no effect in branch-promotion mode (documented via PromoteBranch skipping hooks), `req` serves no purpose here.

Removing it would make the API surface match the actual contract:
```go
func handleAddBranch(ctx context.Context, hctx *HandlerContext, branch string) (*mcp.CallToolResult, error) {
```
Called as `handleAddBranch(ctx, hctx, branch)` from the dispatcher. No behaviour change — just a tighter, more honest signature.

internal/mcp/tool_add.go
lugassawan: Addressed in 7d773c1: updated description to 'applies to task mode only' to accurately reflect that AddPRWorktree always derives its source from PR metadata.

internal/mcp/tool_add.go
lugassawan: Addressed in 7d773c1: added a comment above the switch case explaining that `prNum != 0` intentionally routes negative values to handleAddPR, which re-validates and returns a clear error.

internal/mcp/tool_add_test.go
lugassawan: Addressed in 7d773c1: changed "999" (string) to 999 (integer) so the test exercises the same code path real MCP clients use when sending a JSON number.

internal/mcp/types.go
lugassawan: Addressed in 7d773c1: confirmed intentional — omitempty is the right choice here. Branch-promotion mode returns a minimal result (branch + path only); emitting empty task/source strings would mislead callers. The mode-dependent shape is the contract.

internal/mcp/tool_add.go
lugassawan: Addressed in 7d773c1: removed req from the handleAddBranch signature entirely and updated the dispatch call to handleAddBranch(ctx, hctx, branch).

internal/mcp/tool_add_test.go
```diff
@@ -522,3 +524,326 @@ func TestAddToolTrustGateEnvEscapeHatch(t *testing.T) {
...
+func TestAddPRToolRequiresConfig(t *testing.T) {
+	hctx := &HandlerContext{
+		Runner:   &mockRunner{},
+		Config:   nil,
+		RepoRoot: "/repo",
+		Version:  "test",
+	}
+	handler := handleAdd(hctx)
+
+	result := callTool(t, handler, map[string]any{"pr": "42"})
```
lugassawan: **Medium: `TestAddPRToolRequiresConfig` still passes string `"42"` for `pr` field — same class of bug fixed in `TestAddPRToolFetchError`**

The prior review fixed this exact pattern at line 664 (`"999"` → `999`), but `TestAddPRToolRequiresConfig` at this line was missed. `GetInt` silently coerces strings via `strconv.Atoi`, so the test happens to route to `handleAddPR` and land on the config check — but it uses the wrong type. Every other PR-mode test in this file uses an integer literal.

**Suggested fix:**
```go
result := callTool(t, handler, map[string]any{"pr": 42})
```

internal/mcp/tool_add.go
```diff
@@ -17,94 +17,182 @@ import (
 
 func registerAddTool(s *server.MCPServer, hctx *HandlerContext) {
 	tool := mcp.NewTool("add",
-		mcp.WithDescription("Create a new worktree for a task (supports 'service/task' for monorepo)"),
+		mcp.WithDescription("Create a new worktree for a task, a GitHub PR review, or promote an existing local branch"),
 		mcp.WithString("task",
-			mcp.Description("Task identifier (e.g. 'my-feature', 'JIRA-123', or 'auth-api/my-feature' for monorepo)"),
-			mcp.Required(),
+			mcp.Description("Task identifier (e.g. 'my-feature', 'JIRA-123', or 'auth-api/my-feature' for monorepo); required for normal add, optional as PR name override when pr is set"),
```
lugassawan: **Low: `task` description is silent about branch mode — callers may pass it and be surprised it's ignored**

The description says "optional as PR name override when pr is set" but does not mention that `task` is silently ignored when `branch` is set. An LLM agent that reads the description may pass `{ branch: "feat/x", task: "my-name" }` and expect the name to take effect, when it will not.

Consider appending `; ignored in branch mode` to make the contract explicit.

internal/mcp/tool_add_test.go
lugassawan: Addressed in 2419944b8a5c489c5be005f7636b9e91f6727790: changed "42" to 42 (integer literal) in TestAddPRToolRequiresConfig, matching the same fix applied to TestAddPRToolFetchError.

internal/mcp/tool_add.go
lugassawan: Addressed in 2419944b8a5c489c5be005f7636b9e91f6727790: appended "; ignored in branch mode" to the task field description so callers know the field has no effect when branch is set.

internal/mcp/tool_add_test.go
```diff
@@ -522,3 +524,326 @@ func TestAddToolTrustGateEnvEscapeHatch(t *testing.T) {
...
+	data := unmarshalJSON[addResult](t, result)
+	if data.Branch != "review/42-fix-login-redirect" {
+		t.Errorf("branch = %q, want %q", data.Branch, "review/42-fix-login-redirect")
+	}
+	if !strings.Contains(data.Path, ".worktrees") {
+		t.Errorf("path = %q, expected to contain worktree dir", data.Path)
+	}
+}
+
+func TestAddPRToolNegativeNumber(t *testing.T) {
```
lugassawan, line 590: **Low: `TestAddPRToolNegativeNumber` relies on `testContext` leaving `GH: nil` — passes only because the `prNum <= 0` guard fires before the nil-GH check**

If the guard order in `handleAddPR` were ever swapped (nil-GH check promoted above the range check), this test would silently hit the wrong error branch (`"server startup bug"` instead of `"invalid pr number"`) and continue asserting the wrong message, masking the regression.

Constructing `hctx` with an explicit non-nil `GH` makes the test self-contained and guard-order-independent:

```go
hctx := &HandlerContext{
    Runner:   &mockRunner{},
    GH:       newGhAuthOK(""),
    Config:   testConfig(),
    RepoRoot: "/repo",
    Version:  "test",
}
```

lugassawan: Addressed in a133046d4b155f80a5e6451ec6daeee17e735ce8: replaced testContext (GH=nil) with an explicit HandlerContext using newGhAuthOK("") so the test asserts the range-check error independently of guard ordering.

PR #324, opened 2026-06-13
[feat] Support `rimba --version` as alias for `rimba version`

### LINKED ISSUE #322: [feat] Support rimba --version as an alias for rimba version
## Product framing

**Problem:** Users expect standard CLI conventions where `--version` is a global flag alias for the version subcommand, but rimba currently only supports `rimba version`.

**Value:** Improving UX by supporting the most common way users discover version info (flag-style invocation), matching patterns in nearly all GNU/POSIX CLIs. This is a small quality-of-life win with zero maintenance burden.

**Acceptance criteria:**
- `rimba --version` produces identical output to `rimba version` (full 6-line block)
- No `-v` shorthand (only `--version`)
- Implemented via Cobra's native `rootCmd.Version` + `SetVersionTemplate`
- Existing `version` subcommand remains unchanged

**Impact / Effort:**
- Impact: S — purely additive UX improvement, zero breaking changes
- Effort: S — 5 lines in `cmd/root.go`, no new dependencies (Cobra v1.10.2 already in go.mod)

## Goal
Support `rimba --version` as a global flag alias for `rimba version`, matching standard CLI conventions.

## Motivation
Many users instinctively try `rimba --version` first when checking version info. This is the dominant pattern in POSIX/GNU tools and modern CLIs. Currently rimba only supports the subcommand form (`rimba version`), which requires users to learn an unconventional interface. Supporting both forms costs almost nothing and improves discoverability.

## Surface
- [x] Command (`cmd/*`)
- [ ] Branch / worktree resolver (`internal/resolver/`)
- [ ] Git operations (`internal/git/`)
- [ ] Worktree operations / hooks (`internal/operations/`)
- [ ] MCP server (`internal/mcp/`)
- [ ] Agent files (`internal/agentfile/`)
- [ ] Config (`.rimba/settings.toml`)
- [ ] Documentation
- [ ] Build / CI / release

## Sketch
```bash
# Both should produce identical output:
$ rimba --version
rimba X.Y.Z
commit: abc1234
built:  2026-06-13T12:34:56Z
os:     darwin
arch:   arm64
go:     go1.23.0

$ rimba version
rimba X.Y.Z
commit: abc1234
built:  2026-06-13T12:34:56Z
os:     darwin
arch:   arm64
go:     go1.23.0
```

## Acceptance
- [x] `rimba --version` outputs full 6-line version block (identical to `rimba version`)
- [x] Implementation uses Cobra-native `rootCmd.Version` + `SetVersionTemplate`
- [x] Existing `cmd/version.go` remains unchanged
- [x] No `-v` shorthand; `--version` only
- [x] All existing tests pass

lugassawan:
## Summary

- Extract `versionString()` helper in `cmd/version.go` — single source of truth for both the `version` subcommand and the new `--version` flag
- Wire Cobra's native version flag: set `rootCmd.Version = versionString()` and `SetVersionTemplate("{{.Version}}")` so output is byte-identical to the subcommand
- Pre-register a plain `--version` bool flag to suppress Cobra's auto-added `-v` shorthand (as specified in the issue)
- Output works outside a git repo (Cobra short-circuits before `PersistentPreRunE`)

## Test Plan

- [x] `TestVersionString` — unit test for the extracted helper
- [x] `TestVersionFlagOutput` — asserts `rimba --version` output is byte-identical to `rimba version` subcommand output
- [x] `TestVersionFlagNoShorthand` — asserts `rootCmd.Flags().ShorthandLookup("v")` is nil
- [x] All existing `TestVersion` and `TestVersionCmd` tests continue to pass
- [x] `diff <(rimba version) <(rimba --version)` produces `IDENTICAL`
- [x] Works outside a git repo (no config error)

Closes #322

### DIFF
--- a/cmd/version.go
+++ b/cmd/version.go
@@ -19,13 +19,7 @@ var versionCmd = &cobra.Command{
 	Example:     "  rimba version",
 	Annotations: map[string]string{"skipConfig": "true"},
 	Run: func(cmd *cobra.Command, args []string) {
-		w := cmd.OutOrStdout()
-		fmt.Fprintf(w, "rimba %s\n", version)
-		fmt.Fprintf(w, "commit: %s\n", commit)
-		fmt.Fprintf(w, "built:  %s\n", date)
-		fmt.Fprintf(w, "os:     %s\n", runtime.GOOS)
-		fmt.Fprintf(w, "arch:   %s\n", runtime.GOARCH)
-		fmt.Fprintf(w, "go:     %s\n", runtime.Version())
+		fmt.Fprint(cmd.OutOrStdout(), versionString())
 	},
 }
 
@@ -34,6 +28,20 @@ func Version() string {
 	return version
 }
 
+// versionString returns the multi-line version block printed by both
+// `rimba version` and `rimba --version`.
+func versionString() string {
+	return fmt.Sprintf(
+		"rimba %s\ncommit: %s\nbuilt:  %s\nos:     %s\narch:   %s\ngo:     %s\n",
+		version, commit, date, runtime.GOOS, runtime.GOARCH, runtime.Version(),
+	)
+}
+
 func init() {
 	rootCmd.AddCommand(versionCmd)
+	rootCmd.Version = versionString()
+	rootCmd.SetVersionTemplate("{{.Version}}")
+	// Pre-register --version without a -v shorthand; this prevents Cobra's
+	// InitDefaultVersionFlag from auto-adding -v when the flag is absent.
+	rootCmd.Flags().Bool("version", false, "version for rimba")
 }
--- a/cmd/version.go
+++ b/cmd/version.go
@@ -39,9 +39,8 @@ func versionString() string {
 
 func init() {
 	rootCmd.AddCommand(versionCmd)
-	rootCmd.Version = versionString()
-	rootCmd.SetVersionTemplate("{{.Version}}")
 	// Pre-register --version without a -v shorthand; this prevents Cobra's
 	// InitDefaultVersionFlag from auto-adding -v when the flag is absent.
+	// See TestVersionFlagNoShorthand — if this breaks, Cobra changed InitDefaultVersionFlag.
 	rootCmd.Flags().Bool("version", false, "version for rimba")
 }
--- a/cmd/version.go
+++ b/cmd/version.go
@@ -28,8 +28,6 @@ func Version() string {
 	return version
 }
 
-// versionString returns the multi-line version block printed by both
-// `rimba version` and `rimba --version`.
 func versionString() string {
 	return fmt.Sprintf(
 		"rimba %s\ncommit: %s\nbuilt:  %s\nos:     %s\narch:   %s\ngo:     %s\n",
@@ -39,8 +37,6 @@ func versionString() string {
 
 func init() {
 	rootCmd.AddCommand(versionCmd)
-	// Pre-register --version without a -v shorthand; this prevents Cobra's
-	// InitDefaultVersionFlag from auto-adding -v when the flag is absent.
-	// See TestVersionFlagNoShorthand — if this breaks, Cobra changed InitDefaultVersionFlag.
+	// Pre-register to prevent Cobra's InitDefaultVersionFlag from auto-adding -v; see TestVersionFlagNoShorthand.
 	rootCmd.Flags().Bool("version", false, "version for rimba")
 }
--- a/cmd/version_test.go
+++ b/cmd/version_test.go
@@ -13,8 +13,6 @@ func TestVersion(t *testing.T) {
 	}
 }
 
-// TestVersionString tests versionString() in isolation from I/O — it is the
-// canonical format contract, independent of the subcommand or flag path.
 func TestVersionString(t *testing.T) {
 	got := versionString()
 	for _, want := range []string{"rimba", version, "commit:", "built:", "os:", "arch:", "go:"} {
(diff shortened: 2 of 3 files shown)


cmd/version.go
```diff
@@ -34,6 +28,20 @@ func Version() string {
...
+func versionString() string {
+	return fmt.Sprintf(
+		"rimba %s\ncommit: %s\nbuilt:  %s\nos:     %s\narch:   %s\ngo:     %s\n",
+		version, commit, date, runtime.GOOS, runtime.GOARCH, runtime.Version(),
+	)
+}
+
 func init() {
 	rootCmd.AddCommand(versionCmd)
+	rootCmd.Version = versionString()
```
lugassawan: **Critical** — `rootCmd.Version` is assigned `versionString()` in `init()`. The `-ldflags -X cmd.version=...` mechanism injects values at **link time**, so `version`/`commit`/`date` already hold their final values when `init()` runs — the same values `versionCmd.Run` would see at invocation time. However, if the intent is to ensure the flag always reflects the runtime state (e.g., if `version` were ever updated after start), a deferred assignment in `Execute()` would be more resilient:

```go
func Execute() error {
    updater.SweepOldBinary()
    rootCmd.Version = versionString() // ← post-ldflags, before ExecuteContext
    rootCmd.SetVersionTemplate("{{.Version}}")
    ctx, stop := signal.NotifyContext(context.Background(), os.Interrupt, syscall.SIGTERM)
    defer stop()
    return rootCmd.ExecuteContext(ctx)
}
```

Note: the current implementation is likely correct for release builds (ldflags values are available in `init()`), but moving to `Execute()` removes any ambiguity.

cmd/version_test.go
```diff
@@ -29,3 +42,38 @@ func TestVersionCmd(t *testing.T) {
 		t.Errorf("expected 6 lines, got %d: %q", len(lines), out)
 	}
 }
+
+func TestVersionFlagOutput(t *testing.T) {
+	buf := new(bytes.Buffer)
+	rootCmd.SetOut(buf)
```
lugassawan: **Medium** — `TestVersionFlagOutput` routes both stdout and stderr to the same buffer. If Cobra ever emits anything on stderr (e.g., usage text on error), the byte-equality assertion will fail for a confusing reason. Separate the buffers:

```go
outBuf := new(bytes.Buffer)
errBuf := new(bytes.Buffer)
rootCmd.SetOut(outBuf)
rootCmd.SetErr(errBuf)
// ...
if errBuf.Len() > 0 {
    t.Errorf("unexpected stderr: %q", errBuf.String())
}
if outBuf.String() != want { ... }
```

cmd/version.go
```diff
@@ -34,6 +28,20 @@ func Version() string {
...
+	rootCmd.SetVersionTemplate("{{.Version}}")
+	// Pre-register --version without a -v shorthand; this prevents Cobra's
+	// InitDefaultVersionFlag from auto-adding -v when the flag is absent.
+	rootCmd.Flags().Bool("version", false, "version for rimba")
```
lugassawan, line 41: **Medium** — `rootCmd.Flags().Bool("version", ...)` pre-registers on local flags, which is what Cobra's `InitDefaultVersionFlag` checks (`c.Flags().Lookup("version")`). The mechanism is correct today, but it relies on undocumented Cobra internals. Consider adding a cross-reference comment: `// see TestVersionFlagNoShorthand — if this breaks, Cobra changed InitDefaultVersionFlag` so future Cobra upgrades don't silently regress.

cmd/version_test.go
```diff
@@ -12,10 +13,22 @@ func TestVersion(t *testing.T) {
 	}
 }
 
+func TestVersionString(t *testing.T) {
```
lugassawan, line 16: **Low** — `TestVersionString` and `TestVersionCmd` assert the same 7 substrings + 6-line count. `TestVersionCmd` already exercises the full path (`Run` → `versionString` → write), so `TestVersionString` is only valuable as an isolated unit test for the format contract. If keeping both, add a comment explaining what `TestVersionString` adds beyond `TestVersionCmd` (e.g., that it tests `versionString()` directly without the I/O layer).

cmd/version.go
```diff
@@ -34,6 +28,20 @@ func Version() string {
...
+	return fmt.Sprintf(
+		"rimba %s\ncommit: %s\nbuilt:  %s\nos:     %s\narch:   %s\ngo:     %s\n",
+		version, commit, date, runtime.GOOS, runtime.GOARCH, runtime.Version(),
+	)
+}
+
 func init() {
 	rootCmd.AddCommand(versionCmd)
+	rootCmd.Version = versionString()
+	rootCmd.SetVersionTemplate("{{.Version}}")
```
lugassawan: **Low** — `SetVersionTemplate("{{.Version}}")` works correctly because `versionString()` already includes a trailing `\n`. The template is therefore order-dependent: if `versionString()` ever drops its trailing newline, the version output will run directly into the next shell prompt. Consider making the template self-sufficient: `"{{.Version}}\n"` and drop the trailing `\n` from `versionString()` — or add a test that explicitly asserts the trailing newline.

cmd/version.go
lugassawan: Addressed in dd98670: moved `rootCmd.Version` and `SetVersionTemplate` from `init()` to `Execute()`.

cmd/version_test.go
lugassawan: Addressed in dd98670: split into separate `outBuf`/`errBuf`, added `errBuf.Len() > 0` sanity check.

cmd/version.go
lugassawan: Addressed in dd98670: added cross-reference comment — `// See TestVersionFlagNoShorthand — if this breaks, Cobra changed InitDefaultVersionFlag.`

cmd/version_test.go
lugassawan: Addressed in dd98670: added clarifying comment explaining `TestVersionString` tests the format contract in isolation from the I/O layer.

cmd/version.go
lugassawan: Addressed in dd98670: added `strings.HasSuffix(got, "\n")` assertion in `TestVersionString` to explicitly pin the trailing-newline requirement.

cmd/version_test.go
```diff
@@ -29,3 +47,47 @@ func TestVersionCmd(t *testing.T) {
 		t.Errorf("expected 6 lines, got %d: %q", len(lines), out)
 	}
 }
+
+func TestVersionFlagOutput(t *testing.T) {
```
lugassawan, line 49: **Critical** — `TestVersionFlagOutput` sets `rootCmd.Version` but not `rootCmd.SetVersionTemplate("{{.Version}}")`. The test calls `rootCmd.Execute()` (Cobra's method) directly, bypassing the package-level `Execute()` where the template is now configured. Cobra's default template fires instead: `"{{printf \"%s version %s\\n\" .Name .Version}}"` → output is `"rimba version rimba dev\ncommit: ...\n\n"`, causing the byte-equality assertion to fail.

Fix: mirror the template setup alongside the `Version` setup, with cleanup:
```go
rootCmd.Version = versionString()
rootCmd.SetVersionTemplate("{{.Version}}") // ← add this
...
t.Cleanup(func() {
    rootCmd.Version = origVersion
    rootCmd.SetVersionTemplate("") // restore to Cobra default (empty = use default)
    ...
})
```

lugassawan: Addressed in 46d6586: added `rootCmd.SetVersionTemplate("{{.Version}}")` alongside the existing `rootCmd.Version` setup; cleanup restores to `""` (Cobra default).

PR #327, opened 2026-06-13
[fix] Add Jekyll default layout so docs pages render with the theme

lugassawan:
## Summary
- Add a `defaults` block to `docs/_config.yml` setting `layout: default` for all pages — the root cause of every non-index docs page rendering as bare converted markdown with no theme, no sidebar, and no stylesheet link.
- Add a build-time assertion step in `.github/workflows/pages.yml` that fails CI if representative built pages (init.html, configuration.html, troubleshooting.html) are missing the theme layout wrapper, converting this class of regression into a hard build failure.

## Test Plan
- [ ] Unit tests pass (`make test`) — N/A, no Go files changed
- [ ] Linter passes (`make lint`) — N/A, no Go files changed
- [ ] E2E tests pass (`make test-e2e`) — N/A, no Go files changed

### Type-tailored checks ([fix])
- [x] Reproduce: before the `defaults` block, `bundle exec jekyll build` produced pages with no `<html>` (bare markdown conversion only)
- [x] Verify: after fix, `grep -q '<html' _site/commands/init.html && grep -q 'class="site-header"' _site/commands/init.html` — both match on all three checked pages
- [x] `index.md`'s explicit `layout: home` still wins over the global default (Jekyll front-matter precedence confirmed locally and via reviewer)
- [x] CI guard assertion step passes locally against the fixed build output
- [ ] Post-merge: `curl -sL https://lugassawan.github.io/rimba/commands/init.html | grep -c '<html'` returns `1` after Pages deploy

## Issue

Issue: N/A — site-wide docs rendering regression found via plain-HTML report (all non-index pages missing theme layout due to absent `defaults` block in `_config.yml`)

### DIFF
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -32,11 +32,15 @@ jobs:
       - name: Assert pages render with the theme layout
         working-directory: docs
         run: |
-          for f in _site/commands/init.html _site/configuration.html _site/troubleshooting.html; do
+          for f in _site/commands/init.html _site/commands/add.html \
+                   _site/configuration.html \
+                   _site/troubleshooting.html _site/troubleshooting/adding-promoting.html; do
             [[ -f "$f" ]] || { echo "::error::expected built file $f not found"; exit 1; }
             grep -q '<html' "$f" \
               && grep -q 'class="site-header"' "$f" \
               || { echo "::error file=$f::page rendered without theme layout (missing _config.yml defaults?)"; exit 1; }
+            # class="site-header" is emitted by just-the-docs _layouts/default.html.
+            # If the gem is upgraded, verify this sentinel still matches and update if needed.
           done
           echo "OK: representative pages include the theme layout"
       - uses: actions/upload-pages-artifact@v5
--- a/.github/workflows/pages.yml
+++ b/.github/workflows/pages.yml
@@ -29,6 +29,16 @@ jobs:
         working-directory: docs
         env:
           JEKYLL_ENV: production
+      - name: Assert pages render with the theme layout
+        working-directory: docs
+        run: |
+          for f in _site/commands/init.html _site/configuration.html _site/troubleshooting.html; do
+            [[ -f "$f" ]] || { echo "::error::expected built file $f not found"; exit 1; }
+            grep -q '<html' "$f" \
+              && grep -q 'class="site-header"' "$f" \
+              || { echo "::error file=$f::page rendered without theme layout (missing _config.yml defaults?)"; exit 1; }
+          done
+          echo "OK: representative pages include the theme layout"
       - uses: actions/upload-pages-artifact@v5
         with:
           path: docs/_site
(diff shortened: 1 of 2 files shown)


.github/workflows/pages.yml
```diff
@@ -29,6 +29,16 @@ jobs:
         working-directory: docs
         env:
           JEKYLL_ENV: production
+      - name: Assert pages render with the theme layout
+        working-directory: docs
+        run: |
+          for f in _site/commands/init.html _site/configuration.html _site/troubleshooting.html; do
```
lugassawan: **Medium — spot-check coverage (3 of ~37 pages)**

The loop checks exactly `_site/commands/init.html`, `_site/configuration.html`, and `_site/troubleshooting.html`. The repo has ~37 output pages (25 `commands/` child pages, 8 `troubleshooting/` child pages, etc.).

The `defaults` block is all-or-nothing per Jekyll's design, so any one passing page confirms the global default is active. The regression this guard targets would affect every page simultaneously — spot-checking is sufficient. However, a future change that breaks layout for *only a subtree* (e.g. a per-directory `_config.yml` override or a front-matter `layout:` typo on a child page) would go undetected.

**Suggested fix (optional):** Consider expanding to one representative child page per subtree, e.g. add `_site/commands/worktree.html _site/troubleshooting/common-errors.html` — or accept the current spot-check as good enough given the all-or-nothing nature of `defaults`.

.github/workflows/pages.yml
```diff
@@ -29,6 +29,16 @@ jobs:
...
+            [[ -f "$f" ]] || { echo "::error::expected built file $f not found"; exit 1; }
+            grep -q '<html' "$f" \
+              && grep -q 'class="site-header"' "$f" \
```
lugassawan, line 40: **Medium — `class="site-header"` is an internal theme implementation detail**

The sentinel works correctly today because `just-the-docs` is pinned to exactly `0.12.0` in the Gemfile — the string cannot drift silently on this branch. However, `just-the-docs` has historically renamed structural classes between minor versions (e.g. `site-nav` → `nav-list`), so bumping the gem in future could flip this assertion from a false-pass into a false-fail (CI blocks deploy even though pages render correctly).

**Suggested fix (optional):** When upgrading the theme gem, update this sentinel to match. Alternatively, use a more stable check like `grep -q '<link.*just-the-docs'` on the `<link>` tag pattern (CSS filename also changes on gem bumps, but the `just-the-docs` prefix is stable) — or document this coupling in a comment above the `grep` line so future upgraders know to verify it.

.github/workflows/pages.yml
lugassawan: Addressed in ee5d8c2: expanded loop to also check `_site/commands/add.html` and `_site/troubleshooting/adding-promoting.html` as representative child pages from each subtree.

.github/workflows/pages.yml
lugassawan: Addressed in ee5d8c2: added an inline comment above the `class="site-header"` grep documenting the coupling to just-the-docs `_layouts/default.html` and instructing future gem upgraders to verify or update the sentinel.

PR #336, opened 2026-06-19
[fix] Protect fresh worktrees from clean --merged (#335)

MohammadYusif:
## Summary

- `rimba clean --merged` was force-deleting freshly-created worktrees that had no commits of their own.
- Root cause: `FindMergedCandidates` (`internal/operations/clean.go`) added any branch returned by `git branch --merged <ref>` to the removal set with no own-commits guard. `git branch --merged` lists every branch *reachable* from the ref, including a brand-new worktree branch whose tip *is* the base commit (a trivial ancestor). With the installed `post-merge` hook (which passes `--force`), such fresh worktrees were silently removed and their branches force-deleted.
- The squash-merge fallback (`IsSquashMerged`) already guarded this exact case via `mergeBase == tip`; the plain `--merged` ancestor path did not.
- Fix: added `git.HasOwnCommits` (true only when the branch tip differs from its merge-base with the ref) and wired it into `FindMergedCandidates`, so a `--merged` hit becomes a removal candidate only when the branch actually contributed commits. Fresh and fast-forward-merged branches are protected; squash- and merge-commit-merged branches still get removed. The guard lives in candidate *finding*, not removal, so `--force` cannot bypass it.

## Test Plan

- [x] `make test` — `TestFindMergedCandidates*` pass; new `TestFindMergedCandidatesFreshWorktreeNotRemoved` fails without the guard, passes with it
- [ ] `make lint`
- [ ] `make test-e2e`

Closes #335


internal/git/branch.go
```diff
@@ -131,6 +131,29 @@ func AheadBehind(ctx context.Context, r Runner, dir string) (ahead, behind int,
...
+// HasOwnCommits reports whether branch contains any commits that are not already
+// reachable from mergeRef. It is true when the branch's tip differs from its
+// merge-base with mergeRef. A false result means the branch contributed nothing:
+// either a fresh branch sitting on the base commit, or one that was
+// fast-forward-merged (its tip is an ancestor of mergeRef and adds no new commit).
+//
+// This is used to protect such branches from being treated as "merged" by
+// `git branch --merged`, which lists every branch reachable from the ref —
+// including a brand-new branch whose tip *is* the base commit.
+func HasOwnCommits(ctx context.Context, r Runner, mergeRef, branch string) (bool, error) {
```
lugassawan: **Medium | FF-merged branches with real commits silently survive cleanup**

A branch that was fast-forward merged (e.g. `feature/xyz` at commit C, FF-merged into `main` so both tips equal C) will have `merge-base(main, feature/xyz) == tip == C`. `HasOwnCommits` returns `false`, so it is never added as a removal candidate — even though the work was genuinely merged.

The godoc names this explicitly ("or one that was fast-forward-merged"), framing it as intentional, but users who rely on plain `git merge` (FF allowed) for worktree branches will find those branches silently surviving `rimba clean --merged` forever.

The root tension: `merge-base == tip` cannot distinguish "fresh branch" from "FF-merged branch". Suggestion: document this gap clearly in `clean --help` or a comment in `FindMergedCandidates`, so users are not surprised by stale FF-merged branches persisting.

lugassawan: **Medium | No dedicated unit tests for `HasOwnCommits` in `internal/git`**

`IsSquashMerged` has focused unit tests covering error paths (merge-base error, tip parse, etc.). `HasOwnCommits` is only exercised indirectly through `FindMergedCandidates` — there are no tests in `internal/git/branch_test.go` that exercise it directly.

Given this is new exported API and the project has a 97% coverage threshold, consider adding `TestHasOwnCommits` in `internal/git/branch_test.go` covering:
1. Same SHA → returns `false, nil`
2. Different SHA → returns `true, nil`
3. `MergeBase` error → returns `false, error`
4. `rev-parse` error → returns `false, error`

internal/operations/clean.go
```diff
@@ -69,7 +69,18 @@ func FindMergedCandidates(ctx context.Context, r git.Runner, mergeRef, mainBranc
...
 		if mergedSet[e.Branch] {
-			result.Candidates = append(result.Candidates, CleanCandidate{Path: e.Path, Branch: e.Branch})
+			// `git branch --merged` lists every branch reachable from mergeRef,
+			// including a fresh worktree branch whose tip *is* the base commit.
+			// Guard against removing such branches: only treat a --merged hit as
+			// a candidate when the branch actually contributed commits of its own.
+			hasOwn, err := git.HasOwnCommits(ctx, r, mergeRef, e.Branch)
+			if err != nil {
+				result.Warnings = append(result.Warnings, fmt.Sprintf("skipped %s: own-commits check failed: %v", e.Branch, err))
+				continue
```
lugassawan: **Medium | Missing test for the `HasOwnCommits`-error → warning promotion path**

`TestFindMergedCandidatesSquashMergeError` covers the squash path's error-to-warning promotion. The parallel path added here — `mergedSet[branch] == true` but `HasOwnCommits` fails (e.g. merge-base errors) — converts the error into `result.Warnings` and `continue`s (lines 77-79). This branch has no test.

Suggested addition: `TestFindMergedCandidatesMergedBranchOwnCommitsError` — set up a branch in `mergedSet`, make the mock return an error for `CmdMergeBase`, assert `len(result.Candidates) == 0` and `len(result.Warnings) == 1`. Without it, a regression that accidentally propagates the error instead of warning would not be caught.

internal/git/branch.go
```diff
@@ -131,6 +131,29 @@ func AheadBehind(ctx context.Context, r Runner, dir string) (ahead, behind int,
...
+	mergeBase, err := MergeBase(ctx, r, mergeRef, branch)
+	if err != nil {
+		return false, err
+	}
+
+	tip, err := r.Run(ctx, cmdRevParse, flagVerify, branch)
```
lugassawan: **Low | `rev-parse --verify branch` resolves ambiguously for short branch names**

`HasOwnCommits` passes the bare branch name to `rev-parse --verify branch`. For a branch named the same as a tag, git disambiguates using its refspec priority order, and could resolve to the tag SHA — making the SHA comparison against merge-base unreliable.

`BranchExists` (line 54) consistently uses `refs/heads/<branch>` to be unambiguous. Consider:

```go
tip, err := r.Run(ctx, cmdRevParse, flagVerify, refsHeadsPrefix+branch)
```

(`IsSquashMerged` has the same pre-existing inconsistency, but fixing it there is out of scope for this PR.)

lugassawan (commented):
**Review Decision: COMMENT**

_Reviewed by `reviewer` ([swe-workbench](https://github.com/lugassawan/swe-workbench)). Posted 4 inline comments, deduped 0._

internal/git/branch.go
MohammadYusif: Documented in the `HasOwnCommits` godoc (commit `69462ff`): the comment explicitly notes that `merge-base == tip` cannot distinguish a fresh branch from an FF-merged one, and names this gap. Happy to add a note to `FindMergedCandidates` or `clean --help` too if you'd prefer it surfaced at the user-facing layer.

MohammadYusif: Added in commit `69462ff`: `TestHasOwnCommitsSameSHA` (merge-base == tip → false), `TestHasOwnCommitsDifferentSHA` (different → true), `TestHasOwnCommitsMergeBaseError` (MergeBase error → false, error), and `TestHasOwnCommitsRevParseError` (rev-parse error → false, error). All four cases are in the mock-based unit tests in `internal/git/`.

internal/operations/clean.go
MohammadYusif: Added as `TestFindMergedCandidatesMergedBranchOwnCommitsError` in commit `69462ff`: sets up a branch in `mergedSet`, makes the mock return an error for `MergeBase`, and asserts `len(Candidates) == 0` and `len(Warnings) == 1`.

internal/git/branch.go
MohammadYusif: Already using `refs/heads/<branch>` in commit `69462ff` — the `HasOwnCommits` rev-parse call uses `refsHeadsPrefix+branch` for the same reason `BranchExists` does.

internal/git/branch.go
lugassawan: Upgraded the approach on top of this: `HasOwnCommits` (merge-base == tip) is replaced with `git.IsTipOnFirstParentChain`, which checks whether the branch tip is on `mergeRef`'s first-parent (mainline) history instead of comparing against merge-base. That fixes a related bug the merge-base approach had (it also incorrectly protected merge-commit-merged branches), but the FF-merge trade-off you flagged here is unchanged and still applies — a branch merged via a fast-forward has its tip land on the mainline chain too, so it remains an accepted false-negative. Now formalized with a dedicated test (`TestIsTipOnFirstParentChainFastForward`) and called out in the godoc and PR description rather than left implicit.

lugassawan: The unit tests were renamed/adapted alongside the `HasOwnCommits` → `IsTipOnFirstParentChain` upgrade rather than kept as-is: `internal/git/mock_runner_test.go` now has `TestIsTipOnFirstParentChain{OnChain,OffChain,RevParseError,RevListError,UsesRefsHeadsPrefix}`, and `internal/git/branch_test.go` has real-repo integration tests for the fresh, fast-forward, and (new) merge-commit-merged cases. Coverage on the new function is 100%.

internal/operations/clean.go
lugassawan: This test also carried over: `TestFindMergedCandidatesMergedBranchOwnCommitsError` still exists in `internal/operations/clean_test.go`, now stubbing a `rev-list` failure (the second git call `IsTipOnFirstParentChain` makes) instead of `merge-base`, and still asserts 0 candidates + 1 warning. Also added `TestFindMergedCandidatesMergeCommitRemoved` to lock in the new merge-commit-merged removal case this upgrade fixes.

internal/git/branch.go
lugassawan: Preserved in the upgrade — the `refs/heads/<branch>` prefix carries over unchanged into the new `IsTipOnFirstParentChain` (extracted into a small `BranchTipSHA` helper), and is asserted directly by `TestIsTipOnFirstParentChainUsesRefsHeadsPrefix`.

internal/operations/clean.go
```diff
@@ -139,6 +138,40 @@ func RemoveCandidates(ctx context.Context, r git.Runner, candidates []CleanCandi
...
+}
+
+// classifyMergedEntry decides whether a worktree entry should be removed:
+// the first-parent guard when isMergedHit, else squash-merge detection.
+func classifyMergedEntry(ctx context.Context, r git.Runner, mergeRef string, e git.WorktreeEntry, isMergedHit bool, mainline mainlineLookup) (*CleanCandidate, string) {
+	if isMergedHit {
+		if mainline.err != nil {
+			return nil, fmt.Sprintf("skipped %s: mainline check failed: %v", e.Branch, mainline.err)
+		}
+		tip, err := git.BranchTipSHA(ctx, r, e.Branch)
```
lugassawan: **Medium | Extra `git rev-parse` subprocess per merged-hit worktree**

`classifyMergedEntry` calls `git.BranchTipSHA(ctx, r, e.Branch)` for every merged-hit entry, but `git.WorktreeEntry.HEAD` (already populated by the `ListWorktrees` call earlier in `FindMergedCandidates`) holds the same tip SHA for any branch-backed worktree. This is the same N+1-subprocess pattern the mainline-chain computation was just hoisted out of the loop to fix — leaving this one unhoisted spawns an extra `git rev-parse` per merged-hit worktree on every `clean --merged` run.

Suggested fix: reuse `e.HEAD` directly instead of calling `git.BranchTipSHA`, e.g. `if mainline.shas[e.HEAD] { return nil, "" }`.

internal/git/branch.go
```diff
@@ -131,6 +131,46 @@ func AheadBehind(ctx context.Context, r Runner, dir string) (ahead, behind int,
...
+		if sha := strings.TrimSpace(line); sha != "" {
+			shas[sha] = true
+		}
+	}
+	return shas, nil
+}
+
+// IsTipOnFirstParentChain reports whether branch's tip is on mergeRef's
+// mainline history — true for a fresh or fast-forwarded branch, false for a merge commit.
+func IsTipOnFirstParentChain(ctx context.Context, r Runner, mergeRef, branch string) (bool, error) {
```
lugassawan: **Low | Duplicated first-parent-chain logic between `IsTipOnFirstParentChain` and `classifyMergedEntry`**

`IsTipOnFirstParentChain` is fully implemented and tested but has no production caller — `classifyMergedEntry` in `internal/operations/clean.go` reimplements the same tip-lookup + mainline-membership check inline (to reuse the cached `mainline.shas` set across the loop) rather than delegating to it.

Not a bug today, but a maintenance/drift risk: a future fix to one implementation (e.g. a ref-prefix edge case) could silently miss the other. Worth a short comment noting the duplication is an intentional perf-hoisted path vs. single-check primitive, or extracting a tiny shared `mainline[tip]`-style helper both can call.

internal/operations/clean.go
lugassawan: Addressed in 2cae12b80a0292199438aca499f186b55afd959c: classifyMergedEntry now reuses e.HEAD (already parsed by ListWorktrees) instead of calling git.BranchTipSHA, eliminating the extra rev-parse subprocess.

internal/git/branch.go
lugassawan: Addressed in 2cae12b80a0292199438aca499f186b55afd959c: extracted IsSHAOnChain as the single membership-check primitive, so IsTipOnFirstParentChain and classifyMergedEntry's hoisted mainline-lookup path both delegate to the same implementation instead of duplicating it.

internal/git/branch.go
```diff
@@ -131,6 +131,52 @@ func AheadBehind(ctx context.Context, r Runner, dir string) (ahead, behind int,
 	return ahead, behind, nil
 }
 
+// BranchTipSHA returns the commit SHA that branch currently points to.
+func BranchTipSHA(ctx context.Context, r Runner, branch string) (string, error) {
```
lugassawan: **Medium | `BranchTipSHA` and `IsTipOnFirstParentChain` have zero production callers**

After this round's fix, `classifyMergedEntry` gets the tip from `e.HEAD` and the mainline set from `FirstParentChainSHAs` + `IsSHAOnChain` directly — it no longer calls `IsTipOnFirstParentChain` (or, transitively, `BranchTipSHA`). Both are still exported, documented, and covered by 7 dedicated tests, but represent a code path `clean` no longer exercises in production. `staticcheck`'s `unused` check does not flag exported symbols, so this will not be caught by CI lint, and a future maintainer could reasonably assume `IsTipOnFirstParentChain` is still the canonical entry point.

Suggested fix: remove `BranchTipSHA` and `IsTipOnFirstParentChain` (and their dedicated tests in `branch_test.go`/`mock_runner_test.go`) now that `FirstParentChainSHAs` + `IsSHAOnChain` are the only production-used primitives, or add a comment noting they're currently unused in production if kept for API symmetry.

internal/operations/clean_test.go
```diff
@@ -47,6 +47,13 @@ func TestFindMergedCandidatesNormalMerge(t *testing.T) {
 			if len(args) > 0 && args[0] == gitCmdWorktree {
 				return wt, nil
 			}
+			// IsTipOnFirstParentChain: merge-commit merges leave the tip off mainline.
```
lugassawan: **Low | Misleading comment: describes a code path this test no longer exercises**

The comment `// IsTipOnFirstParentChain: merge-commit merges leave the tip off mainline.` sits above the `rev-parse` stub in `TestFindMergedCandidatesNormalMerge`, but that stub is actually consumed by `IsSquashMerged`'s fallback path for `feature/active` (which is not in the merged set) — `classifyMergedEntry`'s merged-hit path no longer calls `rev-parse` at all (it reads `e.HEAD` directly). Minor, but will mislead anyone debugging this test's mock dispatch later.

Suggested fix: reword to `// IsSquashMerged fallback (feature/active): rev-parse tip for merge-base comparison.`

internal/git/branch.go
lugassawan: Addressed in bbf7ef21f9425689e0711661f70a37857b0d60ef: removed BranchTipSHA and IsTipOnFirstParentChain (no longer called in production). The real-repo integration tests for fresh/merge-commit/fast-forward branches were adapted to exercise FirstParentChainSHAs + IsSHAOnChain directly rather than dropped, and the mock-based unit tests now cover those two primitives instead of the removed wrapper.

internal/operations/clean_test.go
lugassawan: Addressed in bbf7ef21f9425689e0711661f70a37857b0d60ef: reworded the comment to describe what the stub actually does (IsSquashMerged's fallback rev-parse for feature/active), and added a separate comment for the rev-list stub that's actually consumed by classifyMergedEntry's mainline check.

PR #341, opened 2026-06-22
[feat] Support changing branch type (prefix) in rimba rename

lugassawan:
## Summary

- Add `RenameParams` struct to `internal/operations/rename.go` (`NewPrefix ""` = inherit from branch)
- Extend `rimba rename` to accept an optional `[new-task]` arg and prefix flags (`--bugfix`, `--hotfix`, `--docs`, `--test`, `--chore`) so users can change the branch type without recreating the worktree
- Output now shows full branch names (`feature/auth -> bugfix/auth`) instead of task names
- No-op guard updated to `"nothing to change: branch is already <branch>"` (exit 1)

## Test Plan

### Type-tailored checks ([feat])

- [x] `rimba rename auth --bugfix` → `bugfix/auth` branch and `bugfix-auth` dir created; `feature/auth` gone
- [x] `rimba rename auth login --bugfix` → `bugfix/login` (rename + retype in one step)
- [x] `rimba rename auth` (no flag, no change) → exit 1, stderr contains `"nothing to change"`
- [x] Task-only rename still inherits prefix (existing `TestRenamePreservesPrefix` stays green)
- [x] Coverage ≥ 97% (actual: 97.3%)

## Issue

Closes #329


tests/e2e/rename_test.go
```diff
@@ -173,13 +173,68 @@ func TestRenameFailsNoArgs(t *testing.T) {
...
+	r := rimbaSuccess(t, repo, "rename", "retype-src", "retype-dst", "--bugfix")
+	assertContains(t, r.Stdout, "feature/retype-src -> bugfix/retype-dst")
+
+	assertFileNotExists(t, resolver.WorktreePath(wtDir, "feature/retype-src"))
+	assertFileExists(t, resolver.WorktreePath(wtDir, "bugfix/retype-dst"))
+
+	branches := testutil.GitCmd(t, repo, "branch", flagBranchList)
+	if !strings.Contains(branches, "bugfix/retype-dst") {
+		t.Errorf("expected bugfix/retype-dst branch to exist, got:\n%s", branches)
+	}
```
lugassawan, line 256: **Medium:** `TestRenameTaskAndType` does not assert the old branch `feature/retype-src` is gone after the rename. A regression where the old branch is left behind would not be caught.

Suggest adding after line 235:
```go
if strings.Contains(branches, "feature/retype-src") {
    t.Errorf("expected feature/retype-src to be gone after rename, got:\n%s", branches)
}
```

(Consistent with what `TestRenameRetypeOnly` already does.)

docs/commands/rename.md
```diff
@@ -36,16 +38,30 @@ rimba rename auth-changes auth-jwt-migration
...
 | Flag | Description |
 |------|-------------|
 | `-f`, `--force` | Force rename even if the worktree is locked |
 | `--skip-deps` | Skip dependency refresh after rename |
 | `--skip-hooks` | Skip post-rename hooks |
+| `--bugfix` | Change branch type to `bugfix/` |
+| `--hotfix` | Change branch type to `hotfix/` |
+| `--docs` | Change branch type to `docs/` |
+| `--test` | Change branch type to `test/` |
+| `--chore` | Change branch type to `chore/` |
```
lugassawan, line 61: **Low (docs):** The flag table lists 5 non-default prefix flags (`--bugfix`, `--hotfix`, `--docs`, `--test`, `--chore`) but doesn't mention that there is no `--feature` flag. A user who accidentally retypes away from `feature/` cannot self-recover with a flag.

Consider adding a note such as:
> `feature/` is the default prefix — to revert to it, use `rimba rename <task> <task>` (two-arg form with the same task name) which inherits the current prefix without changing anything, then recreate from scratch. Or add a `--feature` flag for symmetry.

tests/e2e/rename_test.go
lugassawan: Addressed in 24d0e28: added `if strings.Contains(branches, "feature/retype-src")` assertion after the branch existence check, consistent with `TestRenameRetypeOnly`.

docs/commands/rename.md
lugassawan: Addressed in 24d0e28: added a note under the flags table clarifying that `feature/` is the default prefix, there is no `--feature` flag, and the recovery path is `rimba remove` + `rimba add` or `git branch -m`.

internal/operations/rename.go
```diff
@@ -17,19 +17,34 @@ type RenameResult struct {
...
+}
+
+// RenameWorktree renames a worktree's directory and branch.
+// It inherits the prefix from the current branch unless p.NewPrefix is set.
+func RenameWorktree(ctx context.Context, r git.Runner, p RenameParams) (RenameResult, error) {
 	prefixes := resolver.AllPrefixes()
 
-	svc, _, matchedPrefix := resolver.ServiceFromBranch(wt.Branch, prefixes)
+	svc, _, matchedPrefix := resolver.ServiceFromBranch(p.WT.Branch, prefixes)
 	if matchedPrefix == "" {
```
lugassawan, line 36: **Medium:** Bare branches (e.g. `my-task`, created with `git branch` directly) silently get promoted to `feature/my-task` on rename because the default-prefix fallback fires here. A user who deliberately named their branch without a prefix will be surprised. Consider documenting this in `docs/commands/rename.md` ("bare branches are promoted to `feature/` on rename") or returning an early error for unrecognized prefixes.

docs/commands/rename.md
```diff
@@ -36,16 +38,32 @@ rimba rename auth-changes auth-jwt-migration
...
+
+> **Note:** There is no `--feature` flag. `feature/` is the default prefix and cannot be selected explicitly. To retype a branch back to `feature/`, use `rimba remove <task>` followed by `rimba add <task>`, or rename the branch directly with `git branch -m`.
```
lugassawan: **Low (docs):** The recovery hint suggests `git branch -m` for retyping back to `feature/`, but `git branch -m` only renames the git branch ref — it does **not** move the worktree directory. After running it, the branch would be `feature/auth` while the directory stays `bugfix-auth`, causing `rimba open` / `rimba merge` to fail (they derive the worktree path from the branch name). Suggest replacing with: "To retype a branch back to `feature/`, use `rimba remove <task>` followed by `rimba add <task>`. Direct `git branch -m` is not recommended because it renames the branch ref but not the worktree directory."

tests/e2e/rename_test.go
```diff
@@ -173,13 +173,71 @@ func TestRenameFailsNoArgs(t *testing.T) {
...
 	if testing.Short() {
 		t.Skip(skipE2E)
 	}
 
+	// 1-arg invocation with no prefix flag — same task, same prefix → no-op error.
 	repo := setupInitializedRepo(t)
-	rimbaFail(t, repo, "rename", "some-task")
+	rimbaSuccess(t, repo, "add", "retype-noop")
+
+	r := rimbaFail(t, repo, "rename", "retype-noop")
```
lugassawan, line 185: **Low:** `TestRenameNoOpReportsCleanly` only covers the default-prefix no-op (`feature/x` with no flag). A user running `rimba rename x --bugfix` when the current branch is already `bugfix/x` would also hit the no-op guard, but this path has no e2e test. The operations-level `TestRenameWorktreeTypeOnly` covers the unit path, but an e2e test would catch command-layer regressions. Suggested addition:
```go
func TestRenameNoOpExplicitSameType(t *testing.T) {
    repo := setupInitializedRepo(t)
    rimbaSuccess(t, repo, "add", "--bugfix", "retype-same")
    r := rimbaFail(t, repo, "rename", "retype-same", "--bugfix")
    assertContains(t, r.Stderr, "nothing to change")
}
```

cmd/rename.go
```diff
@@ -44,9 +49,13 @@ var renameCmd = &cobra.Command{
 			return err
 		}
 
-		_, task = operations.ResolveTaskInput(task, repoRoot)
 		_, newTask = operations.ResolveTaskInput(newTask, repoRoot)
```
lugassawan, line 52: **Medium (advisory):** `ResolveTaskInput` is called on `newTask` here but the source `task` goes through `findWorktree` which internally re-resolves it. The code is functionally correct — monorepo inputs like `auth-api/login` work correctly because `findWorktree` handles the service prefix. The asymmetry is subtle though: a reader of this function must trace into `findWorktree` to understand why `task` doesn't also go through `ResolveTaskInput` at this call site. Consider explicitly resolving both here and passing the resolved task to `findWorktree` so the data flow is auditable in one place. Not a blocker, just a readability note.

internal/operations/rename.go
lugassawan: Addressed in e426b16: added a note to `docs/commands/rename.md`'s intro paragraph documenting that bare branches (without a recognized prefix) are promoted to `feature/` on rename.

docs/commands/rename.md
lugassawan: Addressed in e426b16: removed the `git branch -m` suggestion and replaced it with a note that `git branch -m` is not recommended because it renames the branch ref but not the worktree directory.

tests/e2e/rename_test.go
lugassawan: Addressed in e426b16: added `TestRenameNoOpExplicitSameType` which adds a `bugfix/` worktree and then runs `rimba rename retype-same --bugfix`, asserting the `nothing to change` error.

tests/e2e/rename_test.go
```diff
@@ -173,13 +173,84 @@ func TestRenameFailsNoArgs(t *testing.T) {
...
+func TestRenameNoOpExplicitSameType(t *testing.T) {
+	if testing.Short() {
+		t.Skip(skipE2E)
+	}
+
+	// Explicit --bugfix on an already-bugfix/ branch → no-op error.
+	repo := setupInitializedRepo(t)
+	rimbaSuccess(t, repo, "add", "--bugfix", "retype-same")
+
+	r := rimbaFail(t, repo, "rename", "retype-same", "--bugfix")
```
lugassawan, line 200: **Medium:** Cobra parses `--bugfix` as a flag here (not a second positional arg), so only one arg reaches `RunE`. This is correct, but a reader unfamiliar with cobra flag parsing may count two positional args and conclude the test is malformed (`rename <task> <new-task>` with `--bugfix` as the new-task). A short comment eliminates the ambiguity:
```go
// --bugfix is a flag, not a positional arg; only "retype-same" reaches RunE,
// so newTask == task and the no-op guard fires.
r := rimbaFail(t, repo, "rename", "retype-same", "--bugfix")
```

docs/commands/rename.md
```diff
@@ -36,16 +38,32 @@ rimba rename auth-changes auth-jwt-migration
...
 | `-f`, `--force` | Force rename even if the worktree is locked |
 | `--skip-deps` | Skip dependency refresh after rename |
 | `--skip-hooks` | Skip post-rename hooks |
+| `--bugfix` | Change branch type to `bugfix/` |
+| `--hotfix` | Change branch type to `hotfix/` |
+| `--docs` | Change branch type to `docs/` |
+| `--test` | Change branch type to `test/` |
+| `--chore` | Change branch type to `chore/` |
+
+> **Note:** There is no `--feature` flag. `feature/` is the default prefix and cannot be selected explicitly. To retype a branch back to `feature/`, use `rimba remove <task>` followed by `rimba add <task>`. Direct `git branch -m` is not recommended because it renames the branch ref but not the worktree directory, leaving them out of sync.
```
lugassawan: **Low:** The note correctly warns against `git branch -m` and recommends `rimba remove` + `rimba add`. But the "Retype a misclassified branch" workflow (3 lines above) explicitly says "All uncommitted changes are preserved" — the `remove + add` workaround silently discards them. Consider appending: "Note that `rimba remove` discards uncommitted changes in the worktree; stash or commit first."

tests/e2e/rename_test.go
lugassawan: Addressed in 2d829b4: added a two-line comment explaining that `--bugfix` is a cobra flag (not a positional arg), so `RunE` receives one arg and the no-op guard fires.

docs/commands/rename.md
lugassawan: Addressed in 2d829b4: added "note that this discards uncommitted changes, so stash or commit first" to the `rimba remove + add` recovery path.

tests/e2e/rename_test.go
```diff
@@ -173,13 +173,86 @@ func TestRenameFailsNoArgs(t *testing.T) {
...
+	// Old branch and directory gone
+	assertFileNotExists(t, resolver.WorktreePath(wtDir, "feature/retype-auth"))
+	branches := testutil.GitCmd(t, repo, "branch", flagBranchList)
+	if strings.Contains(branches, "feature/retype-auth") {
+		t.Errorf("expected feature/retype-auth to be gone")
+	}
+
+	// New branch and directory exist
+	assertFileExists(t, resolver.WorktreePath(wtDir, "bugfix/retype-auth"))
+	if !strings.Contains(branches, "bugfix/retype-auth") {
```
lugassawan, line 228: **Medium:** `branches` snapshot is captured at line 220 (before `assertFileExists`) and then used for branch-existence assertions at lines 221–228. The interleaving of a file-check between the snapshot and its branch checks makes the test harder to read and fragile if a future refactor inserts branch-modifying logic between them. `TestRenameTaskAndType` (lines 249–255) does it correctly — `branches` is captured and immediately used for both checks without interleaving.

**Fix:** move `branches := testutil.GitCmd(...)` to after the `assertFileExists` call:
```go
assertFileExists(t, resolver.WorktreePath(wtDir, "bugfix/retype-auth"))
branches := testutil.GitCmd(t, repo, "branch", flagBranchList)
if strings.Contains(branches, "feature/retype-auth") {
    t.Errorf(...)
}
if !strings.Contains(branches, "bugfix/retype-auth") {
    t.Errorf(...)
}
```

docs/commands/rename.md
```diff
@@ -36,16 +38,32 @@ rimba rename auth-changes auth-jwt-migration
...
+> **Note:** There is no `--feature` flag. `feature/` is the default prefix and cannot be selected explicitly. To retype a branch back to `feature/`, use `rimba remove <task>` followed by `rimba add <task>` — note that this discards uncommitted changes, so stash or commit first. Direct `git branch -m` is not recommended because it renames the branch ref but not the worktree directory, leaving them out of sync.
 
 ## Related commands
 
-- [rimba add](add) · create a worktree
-- [rimba duplicate](duplicate) · create a copy with a new name
-- [rimba trust](trust) · approve post-rename shell commands
+- [rimba add]({{ '/docs/commands/add' | relative_url }}) · create a worktree
+- [rimba duplicate]({{ '/docs/commands/duplicate' | relative_url }}) · create a copy with a new name
+- [rimba trust]({{ '/docs/commands/trust' | relative_url }}) · approve post-rename shell commands
```
lugassawan: **Medium:** Related-commands links use the Jekyll `relative_url` filter form (`{{ '/docs/commands/add' | relative_url }}`), but every other page in `docs/commands/` uses bare relative links (`[rimba add](add)`). The project convention for same-level sibling links in Jekyll docs is the bare form; `relative_url` is reserved for cross-level links (e.g. linking from a child page up to a parent).

**Fix:** match the established pattern:
```markdown
- [rimba add](add) · create a worktree
- [rimba duplicate](duplicate) · create a copy with a new name
- [rimba trust](trust) · approve post-rename shell commands
```

tests/e2e/rename_test.go
lugassawan: Addressed in ef1c1f9: moved `branches :=` to after both file-existence assertions, matching the ordering pattern used in `TestRenameTaskAndType`.

docs/commands/rename.md
lugassawan: Addressed in ef1c1f9: changed to bare relative links (`[rimba add](add)`) matching every other sibling page in `docs/commands/`.

cmd/rename_test.go
```diff
@@ -293,3 +293,28 @@ func TestRenameMoveFails(t *testing.T) {
...
+
+func TestRenameRetypeFlag(t *testing.T) {
+	repoDir := t.TempDir()
+	cfg := &config.Config{WorktreeDir: defaultRelativeWtDir, DefaultSource: branchMain}
+	worktreeOut := makeRenameWorktreeOut(repoDir)
+	restore := overrideNewRunner(makeRenameRunner(repoDir, worktreeOut))
+	defer restore()
+
+	cmd, buf := newTestCmd()
+	cmd.Flags().BoolP("force", "f", false, "")
```
lugassawan: **Low:** `cmd.Flags().BoolP("force", "f", false, "")` uses the string literal `"force"` rather than the `flagForce` constant defined in `cmd/flags.go`. Not a correctness issue — cobra returns `false, nil` for an unregistered flag and the force path is not exercised here — but it is inconsistent with how `flagForce` is referenced elsewhere in the file (e.g. `cmd.Flags().GetBool(flagForce)` at line 75 in `rename.go`).

**Fix:** replace the string literal:
```go
cmd.Flags().BoolP(flagForce, "f", false, "")
```

lugassawan: Addressed in a7de919: replaced `"force"` string literal with `flagForce` constant across all 8 occurrences in `cmd/rename_test.go`.

PR #345, opened 2026-06-23
[fix] Block option injection in git worktree add + add subprocess deadline (#330)

lugassawan:
## Summary

- **Security (High):** Add `--` end-of-options separator to `git worktree add` in `internal/git/worktree.go`, preventing a user/config-supplied `source` value beginning with `-` from being parsed by git as a flag
- **Security (High):** Introduce `internal/gitref` package with `Validate()` and `ErrUnsafeRefName` sentinel; validates `default_source` in config at load time and `--source` CLI flag at invocation time, giving a clear error before any subprocess runs
- **Reliability (Medium):** Add `ExecRunner.Timeout` field in `internal/git/git.go` and mirror it in `internal/gh/runner.go`; a `context.WithTimeout` wraps every subprocess call, composing with the existing signal context via earliest-deadline-wins
- **Config:** New `command_timeout` TOML field (e.g. `"30s"`) with `EffectiveCommandTimeout()` accessor; defaults to 120 s; validated in `Config.Validate()`; threaded through all `cmd/` call sites via `newRunner(cmd.Context())` and `newGHRunner(cmd.Context())`
- **Refactor:** `gh.Default(timeout)` now takes an explicit deadline; `newRunner`/`newGHRunner` in `cmd/helpers.go` source the timeout from `config.FromContext(ctx)` in one place

## Test Plan

- [x] Coverage ≥ 97% (`go test ./... -coverprofile=...` → 97.2%)
- [x] `TestAddWorktreeInsertsDashDash` — asserts `--` appears before path/source in args
- [x] `TestExecRunnerTimeoutExpires` — nanosecond timeout causes `context deadline exceeded`
- [x] `TestAddTaskRejectsUnsafeSource` — `--source=-foo` returns "leading dash" error
- [x] `TestValidate` (12 table cases) — all gitref allow/reject paths covered
- [x] `TestEffectiveCommandTimeout`, `TestValidateCommandTimeout`, `TestMergeCommandTimeout`

## Issue

Closes #330

### DIFF
--- a/cmd/add_test.go
+++ b/cmd/add_test.go
@@ -718,6 +718,34 @@ func conflictBranchRunInDirFn(stashDropped *bool) func(dir string, args ...strin
 	}
 }
 
+func TestAddTaskRejectsUnsafeSource(t *testing.T) {
+	cfg := &config.Config{WorktreeDir: "worktrees", DefaultSource: "main"}
+	r := &mockRunner{
+		run:      func(_ ...string) (string, error) { return "", nil },
+		runInDir: noopRunInDir,
+	}
+	restore := overrideNewRunner(r)
+	defer restore()
+
+	cmd, _ := newTestCmd()
+	addPrefixFlags(cmd)
+	_ = cmd.Flags().String(flagSource, "", "")
+	_ = cmd.Flags().String(flagTask, "", "")
+	_ = cmd.Flags().Bool(flagSkipDeps, false, "")
+	_ = cmd.Flags().Bool(flagSkipHooks, false, "")
+	_ = cmd.Flags().SetAnnotation(flagNoColor, cobra.BashCompOneRequiredFlag, nil)
+	cmd.SetContext(config.WithConfig(context.Background(), cfg))
+	_ = cmd.Flags().Set(flagSource, "-foo")
+
+	err := addCmd.RunE(cmd, []string{"my-task"})
+	if err == nil {
+		t.Fatal("expected error for --source=-foo, got nil")
+	}
+	if !strings.Contains(err.Error(), "leading dash") {
+		t.Errorf("error %q should mention 'leading dash'", err)
+	}
+}
+
 func TestAddBranchStashApplyConflict(t *testing.T) {
 	repoDir := t.TempDir()
 	wtDir := filepath.Join(repoDir, "worktrees")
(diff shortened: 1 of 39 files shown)


cmd/add_test.go
```diff
@@ -718,6 +718,34 @@ func conflictBranchRunInDirFn(stashDropped *bool) func(dir string, args ...strin
...
+	restore := overrideNewRunner(r)
+	defer restore()
+
+	cmd, _ := newTestCmd()
+	addPrefixFlags(cmd)
+	_ = cmd.Flags().String(flagSource, "", "")
+	_ = cmd.Flags().String(flagTask, "", "")
+	_ = cmd.Flags().Bool(flagSkipDeps, false, "")
+	_ = cmd.Flags().Bool(flagSkipHooks, false, "")
+	_ = cmd.Flags().SetAnnotation(flagNoColor, cobra.BashCompOneRequiredFlag, nil)
```
lugassawan: **Low:** The `SetAnnotation(flagNoColor, cobra.BashCompOneRequiredFlag, nil)` call marks `--no-color` as a required flag. This is harmless today because the test calls `RunE` directly (bypassing Cobra's required-flag validation), but it's misleading cargo-colt code. Remove the line — `--no-color` needs no annotation.

cmd/update_agent_hint.go
```diff
@@ -40,7 +40,7 @@ func anyInstalled(statuses []agentfile.FileStatus) bool {
 func resolvePostUpdateTipPaths() (string, string) {
 	home, _ := os.UserHomeDir()
 	var repoRoot string
-	if root, err := git.RepoRoot(context.Background(), newRunner()); err == nil {
+	if root, err := git.RepoRoot(context.Background(), newRunner(context.Background())); err == nil {
```
lugassawan: **Low:** This is the one call site that passes `context.Background()` to `newRunner`, meaning a user's `command_timeout` setting is silently ignored here. That's fine for a fire-and-forget tip lookup, but the intent isn't obvious. A short comment explaining the intentional fallback would prevent a future reader from "fixing" it to `cmd.Context()` and discovering there's no `cmd` in scope:

```go
// No cobra cmd here — falls back to DefaultCommandTimeout.
if root, err := git.RepoRoot(context.Background(), newRunner(context.Background())); err == nil {
```

cmd/add_test.go
lugassawan: Addressed in 0d0b55848ebed0e7b6f3bba06e9851aa7e1378f8: removed the spurious `SetAnnotation` call.

cmd/update_agent_hint.go
lugassawan: Addressed in 0d0b55848ebed0e7b6f3bba06e9851aa7e1378f8: added an inline comment documenting the intentional `context.Background()` fallback.

